PR #1, opened 2019-06-21
DEX-265 An independent repository

vsuharnikov:
* Added "future" package for copy-pasted functionality from master;
* Artifact version contains a compatible node's version;
* NODE and other removed projects now included as ProjectRef;
* "matcher" package is renamed to "dex";
* configuration section now is "waves.dex" (was "waves.matcher");
* Updated documentation;
* Removed NODE-related projects;
* Removed unnecessary configuration;


vsuharnikov:
Please use "Create a merge commit" here. We shouldn't lose the history.

PR #12, opened 2019-07-10
DEX-332 Better AskTimeoutException detection

vsuharnikov:
* akka-stream-kafka updated to 1.0.4;

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/api/MatcherApiRoute.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/api/MatcherApiRoute.scala
@@ -4,7 +4,7 @@ import akka.actor.ActorRef
 import akka.http.scaladsl.marshalling.{ToResponseMarshallable, ToResponseMarshaller}
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.{Directive0, Directive1, Route}
-import akka.pattern.ask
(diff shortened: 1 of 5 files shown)


dex/src/main/scala/com/wavesplatform/dex/api/MatcherApiRoute.scala
```diff
@@ -141,12 +141,19 @@ case class MatcherApiRoute(assetPairBuilder: AssetPairBuilder,
         }
     }
 
+  @inline private def askAddressActor(sender: Address, msg: AddressActor.Command): Future[ToResponseMarshallable] =
+    askMapAddressActor[MatcherResponse](sender, msg)(x => x)
```
DenisSerebryansky: Use `identity` function

vsuharnikov: `identity` doesn't work here because of its signature:
```
def identity[A](x: A): A
```
But here we need a function from `MatcherResponse` to `ToResponseMarshallable`. There is an implicit from `MatcherResponse` to `ToResponseMarshallable`, so `x => x` works perfectly.

DenisSerebryansky: Got it

PR #22, opened 2019-07-22
DEX-357 NODE and DEX gRPC interaction

vsuharnikov:
* New Application class;
* Added WavesBlockchainContext - is like Context, but more concrete;
* Matcher depends on WavesBlockchainContext, not on Context;
* Copied several settings from NODE to DEX;
* Removing dependency from Context and Blockchain;


vsuharnikov:
Other fixes will be merged later

PR #25, opened 2019-07-29
DEX-355 gRPC extension for Waves Node

vsuharnikov:


waves-integration-it/src/test/scala/com/wavesplatform/dex/grpc/integration/sync/DEXExtensionTestSuite.scala
```diff
@@ -0,0 +1,153 @@
...
+import com.wavesplatform.transaction.Asset
+import com.wavesplatform.transaction.Asset.Waves
+import io.grpc.stub.StreamObserver
+import io.grpc.{ManagedChannel, ManagedChannelBuilder}
+import mouse.any._
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+
+class DEXExtensionTestSuite extends BaseTransactionSuite {
```
vsuharnikov: `BaseTransactionSuite` is a base class for NODE's integration tests, it doesn't know about Kafka/Local queue. All tests use Kafka queue by default and everyday at night we run all integration tests on a local queue.

So, this test should extend `MatcherSuiteBase`.

DenisSerebryansky: `waves-integration-it` depends on `waves-integration`
`waves-integration` depends on NODE, not DEX

*roll_safe*

PR #39, opened 2019-08-29
DEX-148. Specify tick size in dex

frekensnork:
Cherry-picked and squashed from DEX-148: Specify tick size in DEX
-- Unit- & it-tests added
-- dex-372. Orders on different tick-size levels are matching. It test


dex-it/src/test/scala/com/wavesplatform/it/sync/MatchingRulesTestSuite.scala
```diff
@@ -20,7 +20,27 @@ class MatchingRulesTestSuite extends MatcherSuiteBase {
          |    "$WctId-$UsdId": [
```
frekensnork: here is need in tests for other pairs

PR #58, opened 2019-09-23
DEX-328 reservedBalance of orderV3 is measuring incorrect

frekensnork:
-- it tests


dex-it/src/test/scala/com/wavesplatform/it/sync/OrderFeeTestSuite.scala
```diff
@@ -27,6 +27,14 @@ class OrderFeeTestSuite extends MatcherSuiteBase {
          |    dynamic {
          |      base-fee = $baseFee
          |    }
+         |    percent {
+         |      asset-type = amount
+         |      min-fee = 10
+         |    }
+         |    fixed {
```
frekensnork: there need case in fixed mode paying fee in asset which doesn't take part in trading asset pair

PR #71, opened 2019-10-09
DEX-418 Async order validation

DenisSerebryansky:


dex/src/main/scala/com/wavesplatform/dex/Matcher.scala
```diff
@@ -228,44 +236,44 @@ class Matcher(settings: MatcherSettings, gRPCExtensionClient: DEXClient)(implici
...
+                  val assetPairs: Set[AssetPair] = xs.map { eventWithMeta =>
+                    log.debug(s"Consumed $eventWithMeta")
+                    self ! eventWithMeta
+                    eventWithMeta.event.assetPair
+                  }(collection.breakOut)
+
+                  self
+                    .ask(MatcherActor.PingAll(assetPairs))(pongTimeout)
+                    .recover { case NonFatal(e) => log.error("PingAll is timed out!", e) }
+                    .mapTo[Unit]
```
vsuharnikov: There is an issue. Try to find why :)

PR #80, opened 2019-10-25
DEX-418 Obtaining assets decimals from matching rules during matcher start: errors

vsuharnikov:
Common:
* Removed the context dependency from errors and error formatter;
* Added AssetsDB. An assets' information is saved to the local DB after after getting from the NODE;
* Added com.wavesplatform.dex.effect with FutureResult and utility functions to work with it;
* MatcherSettings contains Asset instead of String where it possible;

An assets' information is obtained:
1. During DEX start for assets in settings.
2. During an order's validation.
3. During a message consumption.

Integration tests:
* Each test specifies its price assets in the config;
* NODE's container now starts before DEX's one, because assets' information is retrieved at start from the NODE;


dex/src/main/scala/com/wavesplatform/dex/Matcher.scala
```diff
@@ -82,8 +88,44 @@ class Matcher(settings: MatcherSettings, gRPCExtensionClient: DEXClient)(implici
...
+                      assetsDB.put(asset, r)
+                      r
+                    }
+                }
+            }
+        }
+    }
+  }
+
+  private lazy val assetsDB: AssetsDB = AssetsDB(db)
```
DenisSerebryansky: What do you think about dedicated class that will incapsulate this logic inside?
Matcher becomes trash class

dex/src/main/scala/com/wavesplatform/dex/Matcher.scala
```diff
@@ -345,34 +395,54 @@ class Matcher(settings: MatcherSettings, gRPCExtensionClient: DEXClient)(implici
...
+          case _                                 => Set.empty
+        }
+      } ++
+        settings.matchingRules.keySet.flatMap(_.assets) ++
+        settings.blacklistedAssets.map(parseAsset("waves.dex.blacklisted-assets"))
+
+      Future
+        .traverse(assetsToLoad) { asset =>
+          getDecimals(assetsDB, wavesBlockchainAsyncClient.assetDescription)(asset).value
+        }
```
DenisSerebryansky: Maybe this also can be incapsulated in the dedicated class

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/clients/async/WavesBlockchainAsyncClient.scala
```diff
@@ -12,6 +17,26 @@ import monix.reactive.Observable
...
+      ))
+
+  final implicit class Ops[F[_]: Functor: Applicative](val self: WavesBlockchainAsyncClient[F]) {
+    def assetDescription(asset: Asset): F[Option[BriefAssetDescription]] =
+      asset.fold(Applicative[F].pure(briefWavesDescription))(self.assetDescription)
+
+    def assetDecimals(asset: Asset.IssuedAsset): F[Option[Int]] = self.assetDescription(asset).map { _.map(_.decimals) }
+
+    // TODO
+    def assetDecimals(asset: Asset): F[Option[Int]] = asset.fold { Applicative[F].pure(Option(8)) } { issuedAsset =>
```
DenisSerebryansky: Use function from line 33?

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/clients/async/WavesBlockchainAsyncClient.scala
```diff
@@ -12,6 +17,26 @@ import monix.reactive.Observable
...
   type SpendableBalance        = Map[Asset, Long]
   type SpendableBalanceChanges = Map[Address, SpendableBalance]
+
+  private val briefWavesDescription =
+    Option(
+      BriefAssetDescription(
+        name = ByteStr("WAVES".getBytes(StandardCharsets.UTF_8)),
+        decimals = 8,
+        hasScript = false
+      ))
```
DenisSerebryansky: Can we find better place for this code?
Looks a bit unnatural

dex/src/main/scala/com/wavesplatform/dex/Matcher.scala
vsuharnikov: Moved to the companion object (changes on the way :) )

dex/src/main/scala/com/wavesplatform/dex/Matcher.scala
vsuharnikov: Probably, we it'd better to move the logic of assets' retrieving from settings. Other lines should stay here IMO, because they depend on incapsulated field of Matcher.

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/clients/async/WavesBlockchainAsyncClient.scala
vsuharnikov: It's used in a case of an issued asset.

DenisSerebryansky: I mean use frunction from line 33 as argument in the second arguments list in `fold`

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/clients/async/WavesBlockchainAsyncClient.scala
vsuharnikov: It is an internal optimization for Ops.assetDescription to avoid unnecessary creation of objects.

DenisSerebryansky: If we place this constant into `BriefAssetDescription` companion object, optimization would not suffer

vsuharnikov: Done

PR #85, opened 2019-10-30
DEX-364 Rates error handling

DenisSerebryansky:
Error handling for the Incorrect updates of the asset rates added


dex/src/main/scala/com/wavesplatform/dex/api/MatcherApiRoute.scala
```diff
@@ -194,19 +196,34 @@ case class MatcherApiRoute(assetPairBuilder: AssetPairBuilder,
...
-        complete(
-          if (asset == Waves) StatusCodes.BadRequest -> wrapMessage("Rate for Waves cannot be changed")
-          else {
-            val assetStr = AssetPair.assetIdStr(asset)
-            rateCache.upsertRate(asset, rate) match {
-              case None     => StatusCodes.Created -> wrapMessage(s"Rate $rate for the asset $assetStr added")
-              case Some(pv) => StatusCodes.OK      -> wrapMessage(s"Rate for the asset $assetStr updated, old value = $pv, new value = $rate")
+  def upsertRate: Route = {
+
+    val rateParsingRejectionsHandler =
```
vsuharnikov: Could you move this outside the `def` to avoid initialization on each call?

DenisSerebryansky: Done

PR #92, opened 2019-11-12
DEX-435 Replace consumer and producer

vsuharnikov:
Dependencies:
* akka-kafka-client is replaced by kafka-clients. We gained more control for Kafka's clients;
* Removed okhttp-backend;
* Resolvers fixed from http to https;
* Updated SBT to 1.3.3;

Settings:
* Removed akka-kafka-client specific settings;
* Added waves.dex.events-queue.kafka.consumer,producer. They contain the "client" section with kafka clients settings;
* Added waves.dex.events-queue.kafka.consumer.fetch-max-duration,max-buffer-size for polling and consumer buffer tuning;

Queues:
* MatcherQueue.lastProcessedOffset is removed. Matcher does this work by itself;
* KafkaMatcherQueue is rewritten to work on kafka-clients instead of akka-kafka-client;

Matcher:
* lastProcessedOffset is a local variable to track last processed offset from the queue;

Tools:
* KafkaTopicInfo is rewritten;

Integration tests:
* CachedData is a helper to skip excess calls to docker client;
* Docker.getInternalSocketAddress,getExternalSocketAddress - repeatable requests if a fail happened;
* Docker.start waits until a container started by checking its logs. This helps to fit a timeout in waitReady on systems where containers started slowly;
* Logs now contains an exception's stacktrace;
* Kafka settings are optimized;


dex/src/main/scala/com/wavesplatform/dex/settings/ConfigExtraOps.scala
```diff
@@ -0,0 +1,15 @@
+package com.wavesplatform.dex.settings
+
+import java.util.Properties
+
+import com.typesafe.config.Config
+
+final class ConfigExtraOps(val self: Config) extends AnyVal {
```
DenisSerebryansky: Please merge this into already existed [ConfigOps](https://github.com/wavesplatform/dex/blob/DEX-355-dex-grpc-separation/dex/src/main/scala/future/com/wavesplatform/settings/utils/ConfigOps.scala)

vsuharnikov: There are several ConfigOps: in NODE project and in DEX.
I can't add this method to the NODE's project. 

In the other side we have a package "future" in DEX that contains ConfigOps. We need to move its contents outside the "future" package, because it was a hack: this package contained a modern functionality from the NODE's project that we couldn't access to (because of a version). So, to use a new functionality we had to just add the "future." prefix into imports. Now:
1. ConfigOps in the NODE doesn't contain all required functionality. It looks like it was removed
2. We will move from ficus to pureconfig to do same things without pain.

So, I created a separate implicit class for general purpose functions.

PR #94, opened 2019-11-14
DEX-378 Errors formatting fixes

vsuharnikov:
MatcherApiRoute: handling client's requests with a wrong JSON ;

MatcherError:
* Introduced error.InvalidJson;
* Amount and Price contain BigDecimal now to properly format their values;

MatchingRule.tickSize now is BigDecimal;

History: Double is replaced by BigDecimal to properly format corner cases;

Unit tests: MatcherApiRouteSpec - added cases with a wrong Order's JSON;

Integration tests: OrderRestrictionsTestSuite - "high amount" now contains a huge amount that improperly formatted before;


dex/src/main/scala/com/wavesplatform/dex/api/MatcherApiRoute.scala
```diff
@@ -19,11 +20,19 @@ import com.wavesplatform.dex._
...
 import com.wavesplatform.dex.error.MatcherError
-import com.wavesplatform.dex.market.MatcherActor.{ForceSaveSnapshots, ForceStartOrderBook, GetMarkets, GetSnapshotOffsets, MarketData, SnapshotOffsetsResponse}
+import com.wavesplatform.dex.market.MatcherActor.{
+  ForceSaveSnapshots,
+  ForceStartOrderBook,
+  GetMarkets,
+  GetSnapshotOffsets,
+  MarketData,
+  SnapshotOffsetsResponse
+}
```
DenisSerebryansky: :angry:

vsuharnikov: This is how scalafmt works :)

PR #95, opened 2019-11-14
DEX-401 Custom further processing of the loading results in gRPC caches

DenisSerebryansky:
Blockchain caches now don't keep failed futures and undesirable results according to the predicate

### DIFF
--- a/waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/caches/BlockchainCache.scala
+++ b/waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/caches/BlockchainCache.scala
@@ -3,23 +3,42 @@ package com.wavesplatform.dex.grpc.integration.caches
 import java.time.Duration
 
 import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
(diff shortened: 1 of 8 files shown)


waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/caches/BlockchainCache.scala
```diff
@@ -3,23 +3,44 @@ package com.wavesplatform.dex.grpc.integration.caches
...
     expiration
       .fold(builder)(builder.expireAfterWrite)
       .build {
-        new CacheLoader[K, V] {
-          override def load(key: K): V = loader(key)
+        new CacheLoader[K, Future[V]] {
+          override def load(key: K): Future[V] = loader(key) andThen {
+            case Failure(exception) => log.error(s"Error while value loading occurred: ", exception); cache.invalidate(key)
+            case Success(value) if invalidationPredicate(value) =>
+              log.warn(s"Discarding value $value for the key $key according to the predicate")
```
vsuharnikov: I think this should be removed. It was useful for testing, but will generate a ton of logs during work.

PR #101, opened 2019-11-18
DEX-370 add checks for getOldestSnapshotOffset to OrderBookSnapshotsTestSuite

ngadiyak:


dex-it/src/test/scala/com/wavesplatform/it/sync/OrderBookSnapshotsTestSuite.scala
```diff
@@ -73,7 +77,12 @@ class OrderBookSnapshotsTestSuite extends MatcherSuiteBase {
   }
 
   "All events are processed after restart" in {
+    val oldestOffset = node.getOldestSnapshotOffset
+
     docker.killAndStartContainer(dockerNodes().head)
+
+    node.getOldestSnapshotOffset should be (oldestOffset)
```
vsuharnikov: Actually we need other tests :)
1. Make snapshots
2. Check all order books have the same offset
3. Restart DEX
4. Verify all order books have the same offset as in 2

PR #112, opened 2019-11-26
DEX version in swagger

ngadiyak:
before changes: we showed the node's version in swagger page
after changes: we show the dex's version


vsuharnikov:
Actually this issue fixed in DEX-355 :)

PR #117, opened 2019-11-27
DEX-494 AssetsStorage and cache

DenisSerebryansky:
* AssetsDB renamed to AssetsStorage
* In AssetsStorage cache implementation added
* levelDB usage switched to cache


dex/src/main/scala/com/wavesplatform/dex/db/AssetsStorage.scala
```diff
@@ -0,0 +1,54 @@
...
+trait AssetsStorage {
+  def put(asset: IssuedAsset, item: BriefAssetDescription): Unit
+  def get(asset: IssuedAsset): Option[BriefAssetDescription]
+}
+
+object AssetsStorage {
+
+  val wavesLifted: FutureResult[BriefAssetDescription] = liftValueAsync { BriefAssetDescription.wavesDescription }
+
+  def cache(db: DB): AssetsStorage = new AssetsStorage {
```
vsuharnikov: Ideologically the cache is a decorator to some interface. 

```scala
def cache(db: DB)
```
doesn't compose well with possible backends (leveldb for this case) and depends on implementation details.

It should be:
```scala
def cache(inner: AssetsStorage): AssetsStorage
```

PR #118, opened 2019-11-28
DEX-496 diffx in tests

DenisSerebryansky:
Diffx:

 * DiffMatcherWithImplicits introduced (with implicits for ByteStrs comparison)
 * It and unit tests now extends DiffMatcherWithImplicits (matchTo using available)

Refactoring:

 * allSnapshotOffsets returns AssetPair instead of String as key in result map
 * Implicit reads/writes/format united in json package object (some moved from DexApi)
 * future.com.wavesplatform package removed
 * ConfigOps and ConfigSettingsValidator moved to dex's settings/utils
 * Only dex's ConfigSettingsValidator now is used
 * ConfigExtraOps removed (using ConfigOps instead)


dex-it-tools/src/main/scala/com/wavesplatform/dex/it/json/package.scala
```diff
@@ -2,14 +2,16 @@ package com.wavesplatform.dex.it
...
+import com.wavesplatform.transaction.assets.exchange.{AssetPair, Order}
 import com.wavesplatform.transaction.{Asset, Transaction, TransactionFactory}
 import play.api.libs.json._
 
 import scala.util.{Failure, Success}
 
 package object json {
-  implicit val byteStr: Format[ByteStr] = Format(
+
+  implicit val byteStrFormat: Format[ByteStr] = Format(
```
vsuharnikov: There can't be `byteStr` which is not `Format` or `Reads` or `Writes` here, so an original name has less factor of duplication:

> byteStr**Format**: **Format**[ByteStr] = **Format**(

vs

> byteStr: **Format**[ByteStr] = **Format**(

IMO an additional word **format** doesn't add value here. Same idea for renamings like `r` (which means "result") to `aLooongNameLikeAFunctionNameButWeKnowWhatFunctionReturns` ( :-) ). Mostly it's just a matter of taste.

But you don't need to change names again.

PR #121, opened 2019-11-29
DEX-520 ErrorInterpolator type classes refactoring

DenisSerebryansky:
* ContextShow switched to cats.Show
* ContextWrites switched to play Writes


vsuharnikov:
Did you check a newly generated documentation? Are there changes?

PR #128, opened 2019-12-10
DEX-530 Exceptions during big block application

vsuharnikov:
Catching exceptions in ExchangeTransactionBroadcastActor and retry, if it is possible by mode

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/market/ExchangeTransactionBroadcastActor.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/market/ExchangeTransactionBroadcastActor.scala
@@ -25,20 +25,26 @@ class ExchangeTransactionBroadcastActor(settings: ExchangeTransactionBroadcastSe
 
   private val default: Receive = {
     case ExchangeTransactionCreated(tx) =>
-      check(tx) match {
+      try check(tx) match {
         case Right(_) => broadcast(List(tx))
         case Left(e)  => logError(tx, e)
+      } catch {
+        case e: Throwable => log.error(s"Can't check or broadcast the transaction ${tx.id()}", e)
       }
   }
 
   private def watching(toCheck: Vector[ExchangeTransaction], next: Vector[ExchangeTransaction]): Receive = {
     case ExchangeTransactionCreated(tx) =>
-      check(tx) match {
+      try check(tx) match {
         case Right(_) =>
           broadcast(List(tx))
           context.become(watching(toCheck, next :+ tx))
 
         case Left(e) => logError(tx, e)
+      } catch {
+        case e: Throwable =>
+          log.warn(s"Can't check or broadcast the transaction ${tx.id()}, will try later", e)
+          context.become(watching(toCheck, next :+ tx))
       }
 
     case Send =>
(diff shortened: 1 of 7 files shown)


dex/src/main/scala/com/wavesplatform/dex/market/ExchangeTransactionBroadcastActor.scala
```diff
@@ -25,20 +25,26 @@ class ExchangeTransactionBroadcastActor(settings: ExchangeTransactionBroadcastSe
...
         case Left(e)  => logError(tx, e)
+      } catch {
+        case e: Throwable => log.error(s"Can't check or broadcast the transaction ${tx.id()}", e)
       }
   }
 
   private def watching(toCheck: Vector[ExchangeTransaction], next: Vector[ExchangeTransaction]): Receive = {
     case ExchangeTransactionCreated(tx) =>
-      check(tx) match {
+      try check(tx) match {
```
DenisSerebryansky: Is it possible to use `Try`?

dex/src/main/scala/com/wavesplatform/dex/market/ExchangeTransactionBroadcastActor.scala
```diff
@@ -48,9 +54,13 @@ class ExchangeTransactionBroadcastActor(settings: ExchangeTransactionBroadcastSe
       val (confirmed, unconfirmed) = toCheck.partition(isConfirmed)
       val (expired, ready)         = unconfirmed.partition(_.timestamp <= expireMs)
 
-      broadcast(ready)
-      log.debug(s"Stats: ${confirmed.size} confirmed, ${ready.size} sent")
-      if (expired.nonEmpty) log.warn(s"${expired.size} failed to send: ${expired.map(_.id().toString).mkString(", ")}")
+      log.debug(s"Stats: ${confirmed.size} confirmed, ${ready.size} ready")
+      if (expired.nonEmpty) log.warn(s"${expired.size} expired: ${expired.map(_.id().toString).mkString(", ")}")
+
+      try broadcast(ready)
```
DenisSerebryansky: The same as above

dex/src/main/scala/com/wavesplatform/dex/market/ExchangeTransactionBroadcastActor.scala
vsuharnikov: Sure, with more ~pain~ code :D 

```scala
Try(check(tx)) match {
  case Success(Right(_)) =>
    broadcast(List(tx))
    context.become(watching(toCheck, next :+ tx))
  
  case Success(Left(e)) => logError(tx, e)
  
  case Failure(e) =>
    log.warn(s"Can't check or broadcast the transaction ${tx.id()}, will try later", e)
    context.become(watching(toCheck, next :+ tx))
}

// or better

Try(check(tx))
  .map {
    case Right(_) => broadcast(List(tx)); true
    case Left(e) => logError(tx, e); false
  }
  .recover { case e => log.warn(s"Can't check or broadcast the transaction ${tx.id()}, will try later", e); true}
  .foreach { watchNext => if (watchNext) context.become(watching(toCheck, next :+ tx)) }
```

The second option is not so bad 🤔

vsuharnikov: I decided to not to do this changes, because ExchangeTransactionBroadcastActor has been totally rewritten in master branch. With `try` keyword we will see less differences between current file and changes in `version-1.1.x`, so merging will be easier.

dex/src/main/scala/com/wavesplatform/dex/market/ExchangeTransactionBroadcastActor.scala
vsuharnikov: This too

PR #129, opened 2019-12-13
DEX-529 DEB should be installed without errors

vsuharnikov:
Configuration:
* Updated organization;
* waves-node-grpc.host is localhost by default;
* Fixed wrong config examples for different networks;
* Improved logback configuration;

DEB installation:
* Added debian scripts for package;

README:
* Improved DEB section of documentation, fixed a wrong configuration;


vsuharnikov:
Integration tests passed expect one. It failed a several merges ago and we have a ticket for it.

PR #131, opened 2019-12-18
DEX-542 Reduce number of artifacts

vsuharnikov:
* Only one artifact for DEX server;
* One ZIP and multiple DEB artifacts for Waves Integration Extension;


vsuharnikov:
Only BroadcastUntilConfirmedTestSuite failed, there is a ticket for it

PR #133, opened 2019-12-19
DEX-541 Broken BroadcastUntilConfirmedTestSuite

vsuharnikov:
* Fixed test suite. We should send requests until NODE will be available;
* Fixed an issue with doubled logs;
* All time in logs now in UTC;

### DIFF
--- a/dex-it/src/test/resources/logback-test.xml
+++ b/dex-it/src/test/resources/logback-test.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-    <property name="pattern" value="%date %-5level [%.25thread] %logger{26} - %msg%n%ex"/>
+    <property name="pattern" value="%date{ISO8601,UTC} %-5level [%.25thread] %logger{26} - %msg%n%ex"/>
 
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <target>System.out</target>
--- a/dex-it/src/test/scala/com/wavesplatform/it/MatcherSuiteBase.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/MatcherSuiteBase.scala
@@ -1,6 +1,7 @@
 package com.wavesplatform.it
 
 import java.net.InetSocketAddress
+import java.time.{LocalDateTime, ZoneId}
 import java.util.concurrent.{Executors, ThreadLocalRandom}
 
 import cats.instances.future.catsStdInstancesForFuture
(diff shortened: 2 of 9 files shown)


dex-it/src/test/resources/logback-test.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-    <property name="pattern" value="%date %-5level [%.25thread] %logger{26} - %msg%n%ex"/>
+    <property name="pattern" value="%date{ISO8601,UTC} %-5level [%.25thread] %logger{26} - %msg%n%ex"/>
```
DenisSerebryansky: Can we somehow log time inside conteiners with UTC+3?

dex-it/src/test/scala/com/wavesplatform/it/MatcherSuiteBase.scala
```diff
@@ -155,7 +156,7 @@ abstract class MatcherSuiteBase
   }
 
   private def print(text: String): Unit = {
-    val formatted = s"---------- $text ----------"
+    val formatted = s"---------- [${LocalDateTime.now(ZoneId.of("UTC"))}] $text ----------"
```
DenisSerebryansky: Or this is the solution for the question above?

dex/src/package/doc/logback.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Add debug="true" to debug configuration -->
 <configuration scan="true" scanPeriod="30 seconds">
-    <property name="logback.include.file" value="${logback.include.file:-./logback.xml}"/>
-    <property name="logback.common.pattern" value="${logback.common.pattern:-%date %-5level [%.25thread] %logger{26} - %msg%n%ex}"/>
+    <property name="logback.include.file" value="${logback.include.file:-./logback-user.xml}"/>
```
DenisSerebryansky: Yet another logback?

dex-it/src/test/resources/logback-test.xml
vsuharnikov: The time in Docker and in test logs had different timezone, and I chose UTC, because we have UTC in all tools we use.

dex-it/src/test/scala/com/wavesplatform/it/MatcherSuiteBase.scala
vsuharnikov: Now we have not only a test name in the logs, but the timestamp too

dex/src/package/doc/logback.xml
vsuharnikov: It's strange to include the file itself :) So, if you want to slightly change the configuration - just add logback-user.xml to the same directory and describe difference in "include" tag.

DenisSerebryansky:
68747470733a2f2f6d656469612e6973746f636b70686f746f2e636f6d2f766563746f72732f617070726f7665642d7374616d702d7369676e2d7365616c2d766563746f722d6964383535303930323030

PR #141, opened 2019-12-30
DEX-561 Long NTP init

vsuharnikov:
* Moved Time and NTP from NODE to DEX;

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/time/NTP.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/time/NTP.scala
@@ -3,8 +3,8 @@ import java.net.{InetAddress, SocketTimeoutException}
 
 import com.wavesplatform.utils.ScorexLogging
 import monix.eval.Task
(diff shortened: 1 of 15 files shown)


dex/src/main/scala/com/wavesplatform/dex/time/NTP.scala
```diff
@@ -0,0 +1,75 @@
...
+        client.open()
+        val info = client.getTime(InetAddress.getByName(ntpServer))
+        info.computeDetails()
+        Option(info.getOffset).map { offset =>
+          val r = if (Math.abs(offset) > offsetPanicThreshold) throw new Exception("Offset is suspiciously large") else offset
+          (info.getAddress, r)
+        }
+      } catch {
+        case _: SocketTimeoutException =>
+          None
```
DenisSerebryansky: Why don't you like oneliners?)

vsuharnikov: Just copied it from Node's repository.

PR #142, opened 2020-01-06
DEX-560 gRPC settings + issues

vsuharnikov:
DEX Server:
* Shutdown cases are handled correctly. Also improved startup;
* Now there are 4 thread pools: broadcast, order history, gRPC and common pool for CPU tasks;
* Scala's global execution context is configurable through config;
* Fixed tasks those minstakenly ran on wrong thread pool;
* MatcherApiRoute.getSettings - supported order versions are calculated everytime;

DEX Server configuration:
* Added waves-blockchain-client settings. They contain gRPC and caches settings;
* grpc.integration.waves-node-grpc is removed;

Integration tests API:
* Added profiling for DEX Container;
* Container's logs contain more system information. Now all logs are wrote through Docker Volume;
* Added HasToxiProxy with ToxiProxyContainer inside. ToxiProxy removed;
* BaseContainersKit now creates the network;
* Increased timeout to start containers both Waves NODE and DEX;
* Removed *ContainerInfo classes. Companion objects of containers include all required for a container creation;

Other changes:
* Added actors logging;
* Fixed AddressActor.getOrdersToCancel;
* Removed not used code.


dex-it-common/src/main/scala/com/wavesplatform/dex/it/config/Implicits.scala
```diff
@@ -0,0 +1,20 @@
...
+        .render(
+          ConfigRenderOptions
+            .concise()
+            .setOriginComments(false)
+            .setComments(false)
+            .setFormatted(true)
+            .setJson(false)
+        )
+  }
+}
```
DenisSerebryansky: Move to _com.wavesplatform.dex.settings.utils.ConfigOps_

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/DexContainer.scala
```diff
@@ -0,0 +1,109 @@
...
+        s"-Dlogback.brief.fullPath=$containerLogsPath/container-$containerName.log",
+        s"-Dlogback.detailed.fullPath=$containerLogsPath/container-$containerName.detailed.log"
+      ).prependIf(isProfilingEnabled) {
+          // https://www.yourkit.com/docs/java/help/startup_options.jsp
+          s"-J-agentpath:/usr/local/YourKit-JavaProfiler-2019.8/bin/linux-x86-64/libyjpagent.so=port=10001,listen=all" +
+            s",sampling,monitors,sessionname=prof-$containerName,snapshot_name_format={sessionname}," +
+            s"dir=$containerLogsPath,logdir=$containerLogsPath,onexit=snapshot"
+        }
+        .mkString(" ", " ", " ")
+  )
```
DenisSerebryansky: Why did you extract this code from _BaseContainer_ (former `create` method) and in fact duplicate it in _DexContainer_ and _WavesNodeContainer_?

dex/src/main/scala/com/wavesplatform/dex/Matcher.scala
```diff
@@ -368,37 +398,50 @@ class Matcher(settings: MatcherSettings, gRPCExtensionClient: DEXClient)(implici
...
-        log.info("Waiting all snapshots are restored ...")
-        waitSnapshotsRestored(settings.snapshotsLoadingTimeout)
-      }
+
       deadline = settings.startEventsProcessingTimeout.fromNow
-      lastOffsetQueue <- {
+      (_, lastOffsetQueue) <- {
+        log.info("Waiting all snapshots are restored ...")
+        waitSnapshotsRestored(settings.snapshotsLoadingTimeout).map { _ =>
+          log.info("All snapshots re restored")
```
DenisSerebryansky: are restored?

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/WavesBlockchainClientBuilder.scala
```diff
@@ -0,0 +1,33 @@
...
+object WavesBlockchainClientBuilder extends ScorexLogging {
+  def async(wavesBlockchainClientSettings: WavesBlockchainClientSettings,
+            monixScheduler: Scheduler,
+            grpcExecutionContext: ExecutionContext): WavesBlockchainClient[Future] = {
+    log.info(s"Building gRPC client for server: ${wavesBlockchainClientSettings.grpc.target}")
+
+    val eventLoopGroup = new NioEventLoopGroup
+    val channel = wavesBlockchainClientSettings.grpc.toNettyChannelBuilder
+      .nameResolverFactory(new DnsNameResolverProvider)
+      .defaultLoadBalancingPolicy("pick_first")
```
DenisSerebryansky: Actually this parameter is used by default

dex-it-common/src/main/scala/com/wavesplatform/dex/it/config/Implicits.scala
vsuharnikov: I can't, it used in `DexContainer` and `WavesNodeContainer`, those are in `dex-it-common`

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/DexContainer.scala
vsuharnikov: 1. It is likely took too much efforts to maintain a common method. It contains a common code, but most arguments to them are differ. Also the code was simple here.

2. It was uncomfortable to edit container's starting options without seeing environment variables (and modifying it too!). This is the reason I moved content of *ContainerInfo classes to related companion objects. There are also a golden rule: keep entities those likely change together in same place. I believe we don't follow it everywhere, but it is a good rule, let's aim it

3. Not big deal, but with different special "create" methods for both containers we get the freedom from any Node's changes. Probably we will use their docker image (from docker hub) in the future. This change won't require a new round of code generalization.

dex/src/main/scala/com/wavesplatform/dex/Matcher.scala
vsuharnikov: Fixed

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/WavesBlockchainClientBuilder.scala
vsuharnikov: Done

dex-it-common/src/main/scala/com/wavesplatform/dex/it/config/Implicits.scala
DenisSerebryansky: `dex-it-common` depends on `dex`, so it should be possible

vsuharnikov: Done

PR #145, opened 2020-01-20
DEX-536 Own force stop application

DenisSerebryansky:
* DEX related stop reasons introduced
 * systemd.service contains corrected RestartPreventExitStatus

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/util/ApplicationStopReason.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/util/ApplicationStopReason.scala
@@ -1,10 +1,7 @@
 package com.wavesplatform.dex.util
 
 sealed abstract class ApplicationStopReason(val code: Int)
-case object Default                  extends ApplicationStopReason(1)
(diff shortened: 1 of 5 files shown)


dex/src/main/scala/com/wavesplatform/dex/util/ApplicationStopReason.scala
```diff
@@ -1,10 +1,7 @@
...
 
 sealed abstract class ApplicationStopReason(val code: Int)
-case object Default                  extends ApplicationStopReason(1)
-case object ErrorStartingMatcher     extends ApplicationStopReason(10)
-case object UnsupportedFeature       extends ApplicationStopReason(38)
-case object PasswordNotSpecified     extends ApplicationStopReason(61)
-case object BaseTargetReachedMaximum extends ApplicationStopReason(72)
-case object FatalDBError             extends ApplicationStopReason(74)
-case object InvalidApiKey            extends ApplicationStopReason(128)
+case object StartingMatcherError    extends ApplicationStopReason(10)
```
vsuharnikov: Could you add where these error codes from?

PR #149, opened 2020-01-21
DEX-573 Update dependencies

vsuharnikov:
Code:

 * ToPbConversions and ToVanillaConversions are renamed to DexToPb/WavesToPbConversions and PbToDex/PbToWavesConversions to have better support in IDE;

Swagger, fixed:

 * X-Api-Key wasn't used in new UI;
 * "Not found definition for Function1RequestContextFutureRouteResult";
 * A path to Order in Swagger's schema;

Build:

 * waves-ext/test included in checkPR, added generators, fixed compilation issues;
 * ExtensionPackaging - all incompatibilities now are shown as warnings, not as an error. All issues should be catch in integration tests;
 * Added RewriteSwaggerConfigPlugin - it patches an original index.html to provide the right configuration;
 * A better way to silence unnecessary warnings in waves-grpc;
 * Added sbt-dependency-check to check libraries for CVEs;
 * Added sbt-updates to check new versions of libraries;
 * Added scalafix dependency to help during migration to new libraries' versions;

Dependencies:

 * Code migrated to a new version: play-json;
 * Other libraries updated and no changes required to the our code;
 * New dependency: "org.webjars" % "swagger-ui" . It contains a fresh UI for Swagger;
 * Not updated: Swagger to 2, Kamon to 2. There are separate tasks;

Integration tests:

 * Added a test that checks we show Swagger UI;
 * Graceful shutdown of containers to preserve all logs;

Other:

 * The com.wavesplatform.dex.grpc.integration directory was cut to com/wavesplatform/and_so_on. It was hard to open files in IDE that have same name in different packages/directories.


vsuharnikov:
It seems the swagger page has been broken... Will solve

project/RewriteSwaggerConfigPlugin.scala
```diff
@@ -0,0 +1,86 @@
...
+    SwaggerUIBundle.presets.apis,
+    SwaggerUIStandalonePreset
+  ],
+  plugins: [
+    SwaggerUIBundle.plugins.DownloadUrl
+  ],
+  layout: "StandaloneLayout",
+  operationsSorter: "alpha"
+});
+window.ui = ui;"""
```
ngadiyak: format

vsuharnikov: There is no `stripMargin`, so the intention was to insert a multiline string to the element (sadly, this won't work because of jsoup). That's why there is no formatting

PR #154, opened 2020-01-31
DEX-565 Basic web socket implementation

DenisSerebryansky:
* MatcherWebSocketRoute introduced
 * Endpoint /ws/time returns current time

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
@@ -1,0 +1,35 @@
+package com.wavesplatform.it.sync.api
+
+import akka.Done
+import akka.actor.ActorSystem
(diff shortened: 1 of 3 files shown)


dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
```diff
@@ -0,0 +1,35 @@
...
+
+    val connected = upgradeResponse.map { upgrade =>
+      // just like a regular http request we can access response status which is available via upgrade.response.status
+      // status code 101 (Switching Protocols) indicates that server support WebSockets
+      if (upgrade.response.status == StatusCodes.SwitchingProtocols) Done
+      else throw new RuntimeException(s"Connection failed: ${upgrade.response.status}")
+    }
+
+    connected.onComplete(println)
+    closed.foreach(_ => println("closed"))
```
vsuharnikov: Hm... Should we block the thread here?

PR #170, opened 2020-02-18
DEX-598 WS for integration tests

ngadiyak:


dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/WebSocket.scala
```diff
@@ -0,0 +1,35 @@
...
+import akka.stream.Materializer
+import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
+import com.wavesplatform.dex.domain.account.KeyPair
+import com.wavesplatform.dex.it.docker.DexContainer
+
+import scala.concurrent.Future
+
+class WebSocket(account: KeyPair, dex: DexContainer) {
+  implicit private val system: ActorSystem        = ActorSystem()
+  implicit private val materializer: Materializer = Materializer.matFromSystem(system)
```
vsuharnikov: 1. Should be outside, because some tests could require multiple WebSocket connections.
2. There is no cleanup for these resources

Consider move system and materializer to HasWebSocket and cleanup in HasWebSocket.afterAll

vsuharnikov: Also HasWebSocket could have a method: `mkWebSocket(account, dex): WebSocket` (we have mkEntities)

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/websockets/HasWebSockets.scala
```diff
@@ -0,0 +1,17 @@
+package com.wavesplatform.dex.it.api.websockets
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.ws.Message
+import akka.stream.Materializer
+
+trait HasWebSockets {
```
vsuharnikov: How about cleanup after all tests?

1. Stop all non stopped web sockets
2. Terminate system

DenisSerebryansky: Done

PR #172, opened 2020-02-20
DEX-621 Market order creation fix

DenisSerebryansky:
* getTradableBalance method fixed
 * IT test added

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
@@ -283,7 +283,7 @@ class AddressActor(owner: Address,
   private def getTradableBalance(forAssets: Set[Asset]): Future[Map[Asset, Long]] = {
     Future
       .traverse(forAssets)(asset => spendableBalance(asset) tupleLeft asset)
-      .map(_.toMap |-| openVolume)
+      .map(xs => (xs.toMap |-| openVolume).withDefaultValue(0L))
   }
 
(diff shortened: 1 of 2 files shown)


dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
```diff
@@ -281,9 +282,10 @@ class AddressActor(owner: Address,
   }
 
   private def getTradableBalance(forAssets: Set[Asset]): Future[Map[Asset, Long]] = {
+    implicit val cleaningGroup: CommutativeMonoid[Map[Asset, Long]] = cats.instances.map.catsKernelStdCommutativeMonoidForMap
     Future
       .traverse(forAssets)(asset => spendableBalance(asset) tupleLeft asset)
-      .map(_.toMap |-| openVolume)
+      .map(_.toMap |+| openVolume.mapValues { -_ })
```
vsuharnikov: What means "we haven't an asset in a tradable balance map" ? This means, that it has the 0 value. The code that represents this statement could be:
```scala
.map(xs => (xs.toMap |-| openVolume).withDefaultValue(0L))
```

This makes the `tradableBalance` function total:

```scala
// PlaceOrder:
def toAcceptedOrder(tradableBalance: Map[Asset, Long]): AcceptedOrder = if (isMarket) MarketOrder(order, tradableBalance) else LimitOrder(order)

// MarketOrder:
def apply(o: Order, tradableBalance: Asset => Long): MarketOrder
```

And this will work in a more general way.

DenisSerebryansky: Sensible

PR #197, opened 2020-03-13
DEX-644 Save order with rounding issues into Postgres history

DenisSerebryansky:
* AddressDirectory isSubmittedFilled condition fixed
 * It tests added

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
@@ -104,6 +104,11 @@ class PostgresHistoryDatabaseTestSuite extends MatcherSuiteBase {
     dex1.api.upsertRate(usd, 0.5)
   }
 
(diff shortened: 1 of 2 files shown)


dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
```diff
@@ -16,12 +16,13 @@ import com.wavesplatform.dex.settings.PostgresConnection._
...
 import io.getquill.{PostgresJdbcContext, SnakeCase}
 import net.ceedubs.ficus.Ficus._
+import org.scalatest.BeforeAndAfterEach
 
 import scala.concurrent.duration.DurationInt
 import scala.io.Source
 import scala.util.Try
 
-class PostgresHistoryDatabaseTestSuite extends MatcherSuiteBase {
+class PostgresHistoryDatabaseTestSuite extends MatcherSuiteBase with BeforeAndAfterEach {
```
ngadiyak: MatcherSuiteBase already has BeforeAndAfterEach

PR #206, opened 2020-03-23
DEX-666 API key in authenticated WS stream

DenisSerebryansky:
* Connection to authenticated balance stream can be done via correct timestamp and signature or X-API-Key. X-API-Key has the highest priority, with correct signature and incorrect key stream connection won't be established
 * WS connection responses have X-Error-Message and X-Error-Code headers
 * withAuth creates simple response or response with aforementioned headers for different types of Routes
 * It-test added

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
@@ -1,15 +1,19 @@
 package com.wavesplatform.it.sync.api
 
+import java.nio.charset.StandardCharsets
+
 import akka.http.scaladsl.model.StatusCodes
 import cats.syntax.option._
+import com.google.common.primitives.Longs
 import com.softwaremill.diffx.{Derived, Diff}
 import com.typesafe.config.{Config, ConfigFactory}
 import com.wavesplatform.dex.api.websockets._
+import com.wavesplatform.dex.api.websockets.headers.{`X-Error-Code`, `X-Error-Message`}
 import com.wavesplatform.dex.domain.asset.Asset
 import com.wavesplatform.dex.domain.asset.Asset.Waves
 import com.wavesplatform.dex.domain.order.OrderType
 import com.wavesplatform.dex.domain.order.OrderType.{BUY, SELL}
-import com.wavesplatform.dex.error.ErrorFormatterContext
+import com.wavesplatform.dex.error.{ApiKeyIsNotValid, ErrorFormatterContext, RequestInvalidSignature}
 import com.wavesplatform.dex.it.api.responses.dex.{OrderStatus => ApiOrderStatus}
 import com.wavesplatform.dex.it.api.websockets.{HasWebSockets, WebSocketAuthenticatedConnection, WebSocketConnection}
 import com.wavesplatform.dex.model.{LimitOrder, OrderStatus}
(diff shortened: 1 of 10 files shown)


dex/src/main/scala/com/wavesplatform/dex/api/MatcherWebSocketRoute.scala
```diff
@@ -74,21 +78,23 @@ case class MatcherWebSocketRoute(addressDirectory: ActorRef, matcher: ActorRef,
...
+          case Success(true) => pass
+          case _             => incorrectSignatureAction
+        }
+    }
 
   /** Requires PublicKey, Timestamp and Signature of [prefix `as`, PublicKey, Timestamp] */
   private def accountUpdates: Route = (path("accountUpdates" / PublicKeyPM) & get) { publicKey =>
-    signedGet("as", publicKey) {
-      handleWebSocketMessages(Flow.fromSinkAndSourceCoupled(Sink.ignore, accountUpdatesSource(publicKey)))
+    val directive = optionalHeaderValueByName(`X-Api-Key`.name).flatMap { maybeKey =>
```
vsuharnikov: Well, the right code here is:
```scala
withAuth | signedGet(balanceStreamPrefix, publicKey) // note, it won't work with in inverse order
```
It reflects the task description: if we have an api key, use it and ignore a signature. Otherwise check the request by a signature.

dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
```diff
@@ -93,10 +97,40 @@ class MatcherWebSocketsTestSuite extends MatcherSuiteBase with HasWebSockets {
 
   "MatcherWebSocketRoute" - {
 
+    "should correctly handle connection rejections" in {
+
+      val kp = mkKeyPair("JIo6cTep_u3_6ocToHa")
+
+      val uriWithoutParams = s"${getBaseBalancesStreamUri(dex1)}${kp.publicKey}"
+      val uriWithParams    = s"$uriWithoutParams?t=${System.currentTimeMillis}&s=incorrect"
```
vsuharnikov: It would be nice to have a test with the right signature too. E.g. interesting case when a signature is right, but an api key is not (it should definitely fail) or vise-versa (pass, because we ignore a signature)

dex/src/main/scala/com/wavesplatform/dex/api/MatcherWebSocketRoute.scala
DenisSerebryansky: Done

dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
DenisSerebryansky: Done

dex-it/src/test/scala/com/wavesplatform/it/sync/api/MatcherWebSocketsTestSuite.scala
```diff
@@ -101,28 +105,45 @@ class MatcherWebSocketsTestSuite extends MatcherSuiteBase with HasWebSockets wit
...
+      ) { (uri, apiKey, expectedStatus, expectedError) =>
         val connection = new WebSocketAuthenticatedConnection(uri, apiKey)
-        Await.result(connection.getConnectionResponse, 1.second).response.status shouldBe expectedStatus
+        val response   = Await.result(connection.getConnectionResponse, 1.second).response
+
+        response.status shouldBe expectedStatus
+
+        expectedError.fold() { error =>
+          response.getHeader(`X-Error-Message`.name).get.value shouldBe error.message.text
+          response.getHeader(`X-Error-Code`.name).get.value shouldBe error.code.toString
```
vsuharnikov: 👍

PR #207, opened 2020-03-24
DEX-638 Add test to MultipleMatchersTestSuite

ngadiyak:


dex-it/src/test/scala/com/wavesplatform/it/MatcherSuiteBase.scala
```diff
@@ -63,4 +68,19 @@ trait MatcherSuiteBase
     stopBaseContainers()
     super.afterAll()
   }
+
+  def createAccountWithBalance(balances: (Long, Asset)*): KeyPair = {
+    val account = KeyPair(ByteStr(s"account-test-${System.currentTimeMillis}".getBytes(StandardCharsets.UTF_8)))
```
vsuharnikov: There should be `${ThreadLocalRandom.current().nextInt()}` instead of `${System.currentTimeMillis}` because the second approach leads to creating one account instead of multiple in several scenarios.

PR #208, opened 2020-03-24
DEX-665 Issued asset balance in WS balance stream

DenisSerebryansky:
Bug:

  Issuing of asset doesn't update asset decimals cache before this asset won't appear in asset pair of any valid order. So WS balance stream cannot denormalize balance by this asset since ErrorFormatterContext (uses asset decimals cache) doesn't know asset's decimals.

 Solution:

  * AddressActor uses getAssetDecimals lambda instead of ErrorFormatterContext for values denormalization. This lambda uses asset decimals cache as main source and gRPC call as secondary one

 Miscellaneous:

  * It test added
  * Matcher minor refactoring

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
@@ -4,13 +4,9 @@ import java.time.{Instant, Duration => JDuration}
 
 import akka.actor.{Actor, ActorRef, Cancellable, Status, Terminated}
 import akka.pattern.{ask, pipe}
-import cats.instances.future.catsStdInstancesForFuture
-import cats.instances.list.catsStdInstancesForList
 import cats.instances.long.catsKernelStdGroupForLong
 import cats.kernel.Group
-import cats.syntax.functor._
 import cats.syntax.group.{catsSyntaxGroup, catsSyntaxSemigroup}
-import cats.syntax.traverse._
 import com.wavesplatform.dex.AddressActor._
 import com.wavesplatform.dex.Matcher.StoreEvent
 import com.wavesplatform.dex.api.CanNotPersist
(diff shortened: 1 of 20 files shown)


dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
```diff
@@ -266,12 +280,15 @@ class AddressActor(owner: Address,
     context.system.scheduler.scheduleOnce(settings.wsMessagesInterval, self, PrepareDiffForWsSubscribers)
   }
 
-  private def mkWsBalances(spendableBalances: Map[Asset, Long]): Map[Asset, WsBalances] = {
+  private def mkWsBalances(spendableBalances: Map[Asset, Long]): Future[Map[Asset, WsBalances]] = {
```
vsuharnikov: Another layer of futures make things harder especially in actor. We have to think:
1. What if it fails
2. What if something happens between we run a future and receive a result

We can fill up `efc` before new assets come to AddressActor. We can do this either in the `Matcher` class, or `SpendableBalanceChanges` actor. I think it easier in `Matcher`, e.g. we can do:

```scala
  private val spendableBalancesActor = {
    actorSystem.actorOf(
      Props(
        new SpendableBalancesActor(
          wavesBlockchainAsyncClient.spendableBalances,
          wavesBlockchainAsyncClient.allAssetsSpendableBalance,
          addressActors
        )
      )
    ) unsafeTap { sba =>
      wavesBlockchainAsyncClient.spendableBalanceChanges
        .mapEval { xs =>
          Task
            .traverse(xs.valuesIterator.flatMap(_.keysIterator)) { asset =>
              Task.fromFuture(getDecimalsFromCache(asset).value)
            }
            .map(_ => xs)
        }
        .foreach(sba ! SpendableBalancesActor.Command.UpdateStates(_))(monixScheduler)
    }
  }
```

Also we could use [bufferIntrospective](https://monix.io/docs/3x/reactive/observable.html#bufferintrospective) with a big size (e.g. 100) and combine maps through monoid to improve the performance of this block because asking DEX extension could take some time.

Note this approach doesn't require `getAssetDecimals` in `AddressActor`.

PR #214, opened 2020-03-30
DEX-678 Address actor dedicated settings

DenisSerebryansky:
* AddressActor got dedicated settings section in application.conf
  * DiffMatcherWithImplicits enriched with string diff (some unit tests fixes)

### DIFF
--- a/dex-test-common/src/main/scala/com/wavesplatform/dex/test/matchers/DiffMatcherWithImplicits.scala
+++ b/dex-test-common/src/main/scala/com/wavesplatform/dex/test/matchers/DiffMatcherWithImplicits.scala
@@ -8,7 +8,7 @@ import org.scalatest.matchers.{MatchResult, Matcher}
 trait DiffMatcherWithImplicits {
 
   implicit val derivedByteStrDiff: Derived[Diff[ByteStr]] = Derived(getDiff[ByteStr](_.toString == _.toString))
-  implicit val derivedStringDiff: Derived[Diff[String]]   = Derived(getDiff[String](_ == _))
+  implicit val derivedStringDiff: Derived[Diff[String]]   = Derived(getDiff[String](_ == _)) // TODO we actually don't need this, find way to remove this
 
   def matchTo[A: Diff](left: A): DiffForMatcher[A] = DiffForMatcher(left)
 }
--- a/dex-test-common/src/main/scala/com/wavesplatform/dex/test/matchers/DiffMatcherWithImplicits.scala
+++ b/dex-test-common/src/main/scala/com/wavesplatform/dex/test/matchers/DiffMatcherWithImplicits.scala
@@ -7,9 +7,8 @@ import org.scalatest.matchers.{MatchResult, Matcher}
 
 trait DiffMatcherWithImplicits {
 
-  private val byteStrDiff: Diff[ByteStr] = getDiff[ByteStr](_.toString == _.toString)
-
-  implicit val derivedByteStrDiff: Derived[Diff[ByteStr]] = Derived(byteStrDiff)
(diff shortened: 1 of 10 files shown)


dex-test-common/src/main/scala/com/wavesplatform/dex/test/matchers/DiffMatcherWithImplicits.scala
```diff
@@ -7,9 +7,8 @@ import org.scalatest.matchers.{MatchResult, Matcher}
 
 trait DiffMatcherWithImplicits {
 
-  private val byteStrDiff: Diff[ByteStr] = getDiff[ByteStr](_.toString == _.toString)
-
-  implicit val derivedByteStrDiff: Derived[Diff[ByteStr]] = Derived(byteStrDiff)
+  implicit val derivedByteStrDiff: Derived[Diff[ByteStr]] = Derived(getDiff[ByteStr](_.toString == _.toString))
+  implicit val derivedStringDiff: Derived[Diff[String]]   = Derived(getDiff[String](_ == _))
```
vsuharnikov: It's really needed? :)

dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
```diff
@@ -564,9 +565,4 @@ object AddressActor {
   private case class PendingCommand(command: OneOrderCommand, client: ActorRef)
 
   private case class InsufficientBalanceOrder(order: Order, insufficientAmount: Long, assetId: Asset)
-
-  final case class Settings(wsMessagesInterval: FiniteDuration, batchCancelTimeout: FiniteDuration, maxActiveOrders: Int)
```
vsuharnikov: Why? Each component has own settings. It's easy to find the settings in a companion object, because it is used only in this class and only in constructor. Also it is a small case class in most cases.

dex-test-common/src/main/scala/com/wavesplatform/dex/test/matchers/DiffMatcherWithImplicits.scala
DenisSerebryansky: Otherwise WebSocketMessagesSerdeSpecification (_string_ status in WsOrder) and MathcerSettingsSpecification (_string_ target in GrpcClientSettings) are broken in misterious way

We haven't seen error messages because they were somehow suppressed, they were just marked as ignored (try it out in master branch!)

dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
DenisSerebryansky: Lets see another point:
Everything we got from application.conf and that is passed through MatcherSettings has it's own dedicated settings

vsuharnikov: MatcherSettings is broken. It *has* all settings, but should only *aggregate* them.

dex-test-common/src/main/scala/com/wavesplatform/dex/test/matchers/DiffMatcherWithImplicits.scala
vsuharnikov: Ok

vsuharnikov: Could you add a comment and TODO ?

DenisSerebryansky: Done

dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
DenisSerebryansky: Done

PR #219, opened 2020-04-06
DEX-683 WS domain messages increment

DenisSerebryansky:
* Update id field has been added to WsAddressState and WsOrderBook. It separately grows for each connection
 * Serialization for authenticated and public stream messages has been updated
 * ActorsWebSocketInteractionsSpecification checks each connection update id growth
 * It tests added


dex/src/main/scala/com/wavesplatform/dex/AddressWsMutableState.scala
```diff
@@ -63,9 +63,26 @@ case class AddressWsMutableState(activeWsConnections: Map[ActorRef, UUID],
     )
   }
 
+  def sendSnapshot(balances: Map[Asset, WsBalances], orders: Seq[WsOrder]): Unit = {
+    pendingWsConnections.keys.foreach(_ ! WsAddressState(balances, orders, 0))
```
vsuharnikov, line 67: Please create WsAddressState before and send it in foreach

dex/src/main/scala/com/wavesplatform/dex/AddressWsMutableState.scala
```diff
@@ -63,9 +63,26 @@ case class AddressWsMutableState(activeWsConnections: Map[ActorRef, UUID],
...
+  }
+
+  def sendDiffs(balances: Map[Asset, WsBalances], orders: Seq[WsOrder]): AddressWsMutableState = copy(
+    activeWsConnections = activeWsConnections.map { // dirty but linear
```
vsuharnikov, line 71: About comment: it wouldn't be a quadratic or worse even we run such loop N times. But we do N-1 more unnecessary work :)

PR #220, opened 2020-04-07
DEX-670 Added tests to NetworkIssuesTestSuite

ngadiyak:


dex-it/src/test/scala/com/wavesplatform/it/sync/networking/NetworkIssuesTestSuite.scala
```diff
@@ -37,8 +38,40 @@ class NetworkIssuesTestSuite extends MatcherSuiteBase with HasToxiProxy {
...
+      wsc.close()
+      dex1.disconnectFromNetwork()
+
+      broadcastAndAwait(mkTransfer(acc, alice.toAddress, 99.waves, Waves))
+      wavesNode1.api.balance(acc.toAddress, Waves) should be (0.999.waves)
+
+      dex1.connectToNetwork()
+
+      dex1.api.tryPlace(mkOrder(acc, wavesUsdPair, SELL, 50.waves, 1.usd)) should failWith (3147270)
+    }
```
DenisSerebryansky, line 58: It's a nice test for checking of DEX's resistance to gRPC connection losses
What do you actually check here? Obtaining balance snapshot via WS API doesn't affect client's ability to place orders

ngadiyak: @vsuharnikov guessed that it can affect and dex could pass the order before obtaining balance changes

vsuharnikov: @DenisSerebryansky it could, please see the issue in JIRA

PR #221, opened 2020-04-09
DEX-691 WS connection lifetime restriction

DenisSerebryansky:
* Connection lifetime is also limited by token (signature) lifetime
 * AuthParams introduced
 * signedGet returns AuthParams

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/PingPongTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/PingPongTestSuite.scala
@@ -36,8 +36,8 @@ class PingPongTestSuite extends MatcherSuiteBase with HasWebSockets {
 
       Thread.sleep(0.1.second)
       eventually {
-        wsac.isClosed shouldBe true
         wsac.getPingsBuffer.size shouldBe 5
+        wsac.isClosed shouldBe true
       }
     }
 
@@ -53,8 +53,8 @@ class PingPongTestSuite extends MatcherSuiteBase with HasWebSockets {
 
         Thread.sleep(0.1.second)
         eventually {
-          wsac.isClosed shouldBe true
           wsac.getPingsBuffer.size should (be >= 3 and be <= 4)
+          wsac.isClosed shouldBe true
         }
       }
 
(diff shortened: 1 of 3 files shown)


dex/src/main/scala/com/wavesplatform/dex/api/MatcherWebSocketRoute.scala
```diff
@@ -107,12 +111,16 @@ case class MatcherWebSocketRoute(addressDirectory: ActorRef,
   private lazy val binaryMessageUnsupportedFailure: Future[PingOrPong]  = mkPongFailure("Binary messages are not supported")
   private def unexpectedMessageFailure(msg: String): Future[PingOrPong] = mkPongFailure(s"Got unexpected message instead of pong: $msg")
 
-  private def createStreamFor(source: Source[TextMessage.Strict, ConnectionSource]): Route = {
+  private def createStreamFor(source: Source[TextMessage.Strict, ConnectionSource], expiration: Option[Long] = None): Route = {
 
     import webSocketSettings._
 
+    val connectionLifetime = expiration.fold(maxConnectionLifetime) { exp =>
+      FiniteDuration(exp - System.currentTimeMillis, TimeUnit.MILLISECONDS).min(maxConnectionLifetime)
```
vsuharnikov, line 119: What if the `exp` will be in the past?

dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/PingPongTestSuite.scala
```diff
@@ -99,17 +99,31 @@ class PingPongTestSuite extends MatcherSuiteBase with HasWebSockets {
...
+
+    "by signed connection lifetime expiration, if it < max-connection-lifetime" in {
+      val wsac = mkWsAuthenticatedConnection(alice, dex1, keepAlive = false, connectionLifetime = 1.5.seconds)
+      wsac.isClosed shouldBe false
+
+      Thread.sleep(1.4.seconds)
+      wsac.isClosed shouldBe false
+
+      Thread.sleep(0.1.second)
+      eventually {
```
ngadiyak, line 123: I think using `eventually` here is not right, because eventually wait for 30 or 40 seconds. it could be false positive

PR #231, opened 2020-04-15
DEX-704 Add test for getting matcher public key from settings

ngadiyak:


dex-it/src/test/scala/com/wavesplatform/it/sync/api/GetSettingsTestSuite.scala
```diff
@@ -18,6 +18,10 @@ class GetSettingsTestSuite extends MatcherSuiteBase {
       dex1.api.settings.networkByte shouldBe 'Y'.toByte
     }
 
+    "return matcher's public key" in {
+      dex1.api.settings.matcherPublicKey should be ("7KNgQbmQj2E5sxuzDQxiXNGLmdkKKY83DHYFJ6iX93Gf")
```
DenisSerebryansky, line 22: I think it could be better to use `matcher.publicKey.toString` instead of hardcode

ngadiyak: ok. fixed

PR #232, opened 2020-04-15
DEX-707 Correct X-User-API-Key requirement

DenisSerebryansky:
* Proxy requests will be rejected if their X-User-Public-Key token isn't match to public key/address from url param
 * Integration tests added


dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/DexApi.scala
```diff
@@ -217,21 +219,24 @@ object DexApi {
...
-          .headers(apiKeyHeaders)
-          .headers(xUserPublicKey.fold(Map.empty[String, String])(userPublicKeyHeaders))
-          .contentType("application/json", "UTF-8")
-      }
+      override def tryCancelWithApiKey(id: Order.Id, xUserPublicKey: Option[PublicKey]): F[Either[MatcherError, MatcherStatusResponse]] =
+        tryParseJson {
+          sttp
+            .post(uri"$apiUri/orders/cancel/${id.toString}")
+            .headers(apiKeyHeaders)
+            .headers(xUserPublicKey.fold(Map.empty[String, String])(userPublicKeyHeaders))
```
vsuharnikov, line 237: `apiKeyWithUserPublicKeyHeaders(xUserPublicKey)` ?

PR #236, opened 2020-04-20
DEX-711 WebSocketPrivateStreamTestSuite stabilization

DenisSerebryansky:


dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WebSocketPrivateStreamTestSuite.scala
```diff
@@ -77,7 +77,7 @@ class WebSocketPrivateStreamTestSuite extends MatcherSuiteBase with HasWebSocket
 
       "when user places and cancels limit order" in {
 
-        val acc = mkAccountWithBalance(150.usd -> usd, 10.waves -> Waves)
+        val acc = mkAccountWithBalance(150.usd -> usd, 10.waves -> Waves); Thread.sleep(150)
```
ngadiyak, line 80: it will be fixed in DEX-690

PR #245, opened 2020-04-24
DEX-681 Closing all WS connections after order book deleting

DenisSerebryansky:
* AggregatedOrderBookActor closes all WS connections on PostStop signal
 * ConnectionSource removed, typed ActorRef is used instead
 * TerminationStatus removed, MatcherError is used instead
 * WsConnectionPongTimeout and WsConnectionMaxLifetimeExceeded matcher errors introduced
 * Used typed actor as a recipient of changes in AddressActor
 * ActorsWebSocketInteractionsSpecification rewritten in typed ActorRef style
 * WS connections now have a UUID in names
 * Integration test added

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/market/AggregatedOrderBookActor.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/market/AggregatedOrderBookActor.scala
@@ -5,7 +5,7 @@ import akka.actor.typed.scaladsl.Behaviors
 import akka.actor.typed.{ActorRef, Behavior, Terminated}
 import akka.http.scaladsl.model.{ContentTypes, HttpEntity, HttpResponse}
 import com.wavesplatform.dex.OrderBookWsState
-import com.wavesplatform.dex.api.websockets.WsOrderBook
+import com.wavesplatform.dex.api.websockets.{WsMessage, WsOrderBook}
 import com.wavesplatform.dex.domain.asset.AssetPair
 import com.wavesplatform.dex.domain.model.{Amount, Price}
 import com.wavesplatform.dex.market.OrderBookActor.MarketStatus
@@ -31,6 +31,7 @@ object AggregatedOrderBookActor {
   object Command {
     case class ApplyChanges(levelChanges: LevelAmounts, lastTrade: Option[LastTrade], ts: Long) extends Command
     case class AddWsSubscription(client: ActorRef[WsOrderBook])                                 extends Command
+    case class CloseAllWsSubscriptions(reason: Throwable)                                       extends Command
     private[AggregatedOrderBookActor] case object SendWsUpdates                                 extends Command
   }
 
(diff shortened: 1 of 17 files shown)


dex/src/main/scala/com/wavesplatform/dex/market/AggregatedOrderBookActor.scala
```diff
@@ -27,11 +28,12 @@ object AggregatedOrderBookActor {
...
   }
 
-  sealed trait Command extends Message
-  object Command {
-    case class ApplyChanges(levelChanges: LevelAmounts, lastTrade: Option[LastTrade], ts: Long) extends Command
-    case class AddWsSubscription(client: ActorRef[WsOrderBook])                                 extends Command
-    private[AggregatedOrderBookActor] case object SendWsUpdates                                 extends Command
+  sealed trait WsCommand extends Message
+  object WsCommand {
+    case class ApplyChanges(levelChanges: LevelAmounts, lastTrade: Option[LastTrade], ts: Long) extends WsCommand
```
vsuharnikov, line 33: 1. Actually, it is not a `WsCommand`. This is a regular order book update. An aggregated order book is used on REST API too.
2. Also see no reason to introduce a special `WsCommand` (Occam's razor)

dex/src/main/scala/com/wavesplatform/dex/market/OrderBookActor.scala
```diff
@@ -113,14 +114,9 @@ class OrderBookActor(settings: Settings,
...
-              // DEX-681
-              // private val reason = new EOFException("Order book was deleted")
-              // reason.setStackTrace(Array.empty)
-              // private val wsCloseMessage = classic.Status.Failure(reason)
-              //
-              // wsState.wsConnections.foreach(_ ! wsCloseMessage)
-              // wsState = wsState.withoutSubscriptions
+            case _: QueueEvent.OrderBookDeleted =>
+              log.warn("Order book was deleted, closing all WebSocket connections...")
+              aggregatedRef ! AggregatedOrderBookActor.WsCommand.CloseAllWsSubscriptions(OrderBookStopped(assetPair))
```
vsuharnikov, line 119: Probably we don't need a special message to stop child. See https://doc.akka.io/docs/akka/current/typed/guide/tutorial_1.html#the-actor-lifecycle and try to write something like:
```scala
// AggregatedOrderBookActor
.receiveSignal {
  // ...
  case (_, PostStop) => // here the code
}
```

dex/src/main/scala/com/wavesplatform/dex/market/AggregatedOrderBookActor.scala
DenisSerebryansky: Done

dex/src/main/scala/com/wavesplatform/dex/market/OrderBookActor.scala
DenisSerebryansky: Done

PR #246, opened 2020-04-27
DEX-719 Remove UUID from Ping/Pong

vsuharnikov:


dex/src/main/scala/com/wavesplatform/dex/api/websockets/actors/SystemMessagesHandlerActor.scala
```diff
@@ -87,22 +84,21 @@ object SystemMessagesHandlerActor {
...
 
   object PingOrPong {
 
-    def wsUnapply(arg: PingOrPong): Option[(String, UUID, Long)] = (arg.tpe, arg.connectionId, arg.timestamp).some
+    def wsUnapply(arg: PingOrPong): Option[(String, Long)] = (arg.tpe, arg.timestamp).some
 
     implicit val format: Format[PingOrPong] = (
       (__ \ "T").format[String] and
-        (__ \ "_").format[Long] and
-        (__ \ "%").format[String]
```
DenisSerebryansky, line 102: Do you think we won't need this?

vsuharnikov: Yes, because we have the timestamp

PR #247, opened 2020-04-27
DEX-620 Add the closedOnly query parameter to order history methods

vsuharnikov:


dex/src/main/scala/com/wavesplatform/dex/AddressActor.scala
```diff
@@ -551,6 +553,14 @@ object AddressActor {
 
   private case class InsufficientBalanceOrder(order: Order, insufficientAmount: Long, assetId: Asset)
 
+  sealed abstract class OrderListType(val hasActive: Boolean, val hasClosed: Boolean) extends Product with Serializable
+  object OrderListType {
+    case object All        extends OrderListType(true, true)
+    case object Empty      extends OrderListType(false, false)
```
DenisSerebryansky, line 559: image

vsuharnikov: Yeah, it is possible :)

PR #248, opened 2020-04-28
DEX-717 Real-time spendable balance changes stream

DenisSerebryansky:
* Real-time spendable balance changes stream introduced. DEX extension now tracks balances of all the addresses. It sends only changes that wasn't sent to DEX before
 * Batched stream of spendable balance changes was deprecated, using real-time stream instead. Batched stream should be removed after 2.1.2 release
 * spendableBalance gRPC method removed, using spendableBalances instead
 * WebSocketPrivateStreamTestSuite fixed since now DEX doesn't receive already consumed balance changes from Node


waves-ext/src/main/scala/com/wavesplatform/dex/grpc/integration/services/WavesBlockchainApiGrpcService.scala
```diff
@@ -33,20 +34,57 @@ import scala.concurrent.Future
...
+    realTimeBalanceChangesSubscribers.forEach { _.onError(shutdownError) }
+    realTimeBalanceChangesSubscribers.clear()
   }
 
+  private val realTimeBalanceChanges: Coeval[CancelableFuture[Unit]] = Coeval.evalOnce {
+    context.spendableBalanceChanged
+      .map {
+        case (address, asset) =>
+          val newAssetBalance     = context.utx.spendableBalance(address, asset)
+          val maybeAddressBalance = Option(allSpendableBalances get address)
```
vsuharnikov, line 68: How about
```scala
Option(allSpendableBalances get address).getOrElse(Map.empty)
```
? This will make the code below easier

DenisSerebryansky: Fixed

PR #252, opened 2020-05-03
DEX-720 Order book settings in public WS API stream

DenisSerebryansky:
* Order book settings are sent in snapshot
   - WsOrderBookSettings introduced
   - Unit and integration tests added
   - Using of OrderRestrictionsSettings in OrderValidator refactored
 * Changes of tick-size are sent to WS API public stream
   - Using lenses for modifying of AggregatedOrderBookActor state
 * Unit and integration tests added

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/OrderBookWsState.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/OrderBookWsState.scala
@@ -1,39 +1,33 @@
 package com.wavesplatform.dex
 
 import akka.actor.typed.ActorRef
+import cats.syntax.option._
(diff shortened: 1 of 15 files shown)


dex/src/main/scala/com/wavesplatform/dex/OrderBookWsState.scala
```diff
@@ -80,4 +77,14 @@ case class OrderBookWsState(wsConnections: Map[ActorRef[WsOrderBook], Long],
     }
     r.result()
   }
+
+  def accumulateChanges(lc: LevelAmounts, lt: Option[LastTrade], ts: Option[Double]): OrderBookWsState =
```
vsuharnikov, line 81: 👍

PR #256, opened 2020-05-06
DEX-699 Errors during WebSocket connection

vsuharnikov:
* WsHandlerActor the closing logic changed, no deadletter messages now;
* AggregatedOrderBookActor - returned a previous way to notify clients about removed order book. The client will not receive a message about a deleted order book when the server is stopping;
* WsPingPongTestSuite: now checks that a client receives WsConnectionMaxLifetimeExceeded and WsConnectionPongTimeout before the connection is closed due to one of this reasons;

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsPingPongTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsPingPongTestSuite.scala
@@ -1,11 +1,12 @@
 package com.wavesplatform.it.sync.api.ws
 
 import com.typesafe.config.{Config, ConfigFactory}
+import com.wavesplatform.dex.api.websockets.WsError
 import com.wavesplatform.it.WsSuiteBase
 
 import scala.concurrent.duration._
 
-class PingPongTestSuite extends WsSuiteBase {
+class WsPingPongTestSuite extends WsSuiteBase {
 
   private val maxConnectionLifetime = 6.seconds
   private val pingInterval          = 1.second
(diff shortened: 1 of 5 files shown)


dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsPingPongTestSuite.scala
```diff
@@ -35,9 +38,16 @@ class PingPongTestSuite extends WsSuiteBase {
...
       Thread.sleep(0.1.second)
       eventually {
-        wsac.pings.size shouldBe 5
+        wsac.pings.size should be >= 5
         wsac.isClosed shouldBe true
       }
+
+      wsac.collectMessages[WsError].head.copy(timestamp = 0L) should matchTo(
+        WsError(
+          timestamp = 0L,
```
DenisSerebryansky, line 47: Use ignoring of `timestamp` field

dex/src/main/scala/com/wavesplatform/dex/api/websockets/actors/WsHandlerActor.scala
```diff
@@ -130,22 +125,19 @@ object WsHandlerActor {
...
-              context.log.debug("Client actor stopped, stopping...")
+              cancelSchedules(nextPing, pongTimeout)
+              Behaviors.same // Will receive Completed when WsServerMessage.Complete will be delivered
+
+            case Completed(status) =>
+              status match {
+                case Left(e)  => context.log.debug("Got failure Completed. Stopping...", e)
+                case Right(_) => context.log.debug("Got successful Completed. Stopping...")
+              }
+              cancelSchedules(nextPing, pongTimeout)
```
DenisSerebryansky, line 140: Lines 132 and 140 - only one should survive

vsuharnikov: `Completed` could be even there was no `CloseConnection` command.
But when we receive `CloseConnection` we should not wait and cancel schedules immediately.

PR #259, opened 2020-05-07
DEX-633 Tests backward compatibility 

ngadiyak:


vsuharnikov:
Moved to https://github.com/wavesplatform/matcher/pull/270 because it has the right branch name

PR #268, opened 2020-05-13
DEX-731 WS connection ID

DenisSerebryansky:
Main:
  * WsInitial server message introduced. It contains connection id and is sent to the client right after connection creation
  * WsError is created with time.getTimestamp() in any places
  * Unit and integration tests are fixed

 Tests:
  * mkWsConnection automatically checks WsInitial receiving

 Minor improvements:
  * Completed message in WsHandlerActor and OrderBookDeleted message in AggregatedOrderBookActor became Event
  * WsServerMessage got abstract timestamp field
  * Formats of some ws messages and entities got more explicit names (useful for debugging)

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsPingPongTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsPingPongTestSuite.scala
@@ -34,10 +34,10 @@ class WsPingPongTestSuite extends WsSuiteBase {
       val wsac = mkWsAddressConnection(alice, dex1)
       wsac.isClosed shouldBe false
 
-      Thread.sleep(maxConnectionLifetime - 0.1.second)
+      Thread.sleep(maxConnectionLifetime - 0.2.second)
       wsac.isClosed shouldBe false
 
-      Thread.sleep(0.1.second)
+      Thread.sleep(0.2.second)
       eventually {
         wsac.pings.size should be >= 5
         wsac.isClosed shouldBe true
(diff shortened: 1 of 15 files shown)


dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsPingPongTestSuite.scala
```diff
@@ -115,14 +115,14 @@ class WsPingPongTestSuite extends WsSuiteBase {
       wsac1.send(wsac2.pings.head) // send correct pong but from another connection
       wsac2.send(wsac1.pings.head) // send correct pong but from another connection
 
-      Thread.sleep(pongTimeout - 0.2.second)
+      Thread.sleep(pongTimeout - 0.3.second)
```
ngadiyak, line 118: why?

DenisSerebryansky: Additional awaiting of WsInitial in `mkWsConnection` affects test

dex/src/main/scala/com/wavesplatform/dex/api/websockets/actors/WsHandlerActor.scala
```diff
@@ -52,13 +55,16 @@ object WsHandlerActor {
...
             clientRef: ActorRef[WsServerMessage],
             matcherRef: classic.ActorRef,
-            addressRef: classic.ActorRef): Behavior[Message] =
+            addressRef: classic.ActorRef,
+            connectionId: String): Behavior[Message] =
     Behaviors.setup[Message] { context =>
       import context.executionContext
       import settings.subscriptions._
 
       context.setLoggerName(s"WsHandlerActor[c=${clientRef.path.name}]")
```
vsuharnikov, line 64: Could be:
```scala
context.setLoggerName(s"WsHandlerActor[c=$connectionId]")
```

PR #274, opened 2020-05-19
DEX-751 WS API check

DenisSerebryansky:
Main:
  * DexWsConnector and AuthServiceRestConnector introduced
  * During initialization SuperConnector establishes connection to:
    - DEX WS API and awaits WsInitial message
    - Auth Service REST API and tries to load swagger UI page
  * Check of the order book WS stream added
  * Check of the account updates WS stream added. If Auth Service REST API uri was not specified, Checker skips account updates check
  * Checker can test account updates for the specified seed

Miscellaneous:
  * WsConnection, WsConnectionOps, TestWsHandlerActor moved to dex module
  * Full WS uri passed to WsConnection (prefix is set from outside)
  * repeatRequest and repeatRequestOptions moved from RestConnector to Connector

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/tool/connectors/AuthServiceRestConnector.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/tool/connectors/AuthServiceRestConnector.scala
@@ -1,0 +1,55 @@
+package com.wavesplatform.dex.tool.connectors
+
+import java.nio.charset.StandardCharsets
+import java.util.concurrent.ThreadLocalRandom
+
+import cats.syntax.either._
+import com.wavesplatform.dex.auth.JwtUtils
+import com.wavesplatform.dex.domain.account.KeyPair
+import com.wavesplatform.dex.domain.crypto
+import com.wavesplatform.dex.tool._
+import com.wavesplatform.dex.tool.connectors.AuthServiceRestConnector.AuthCredentials
+import sttp.client._
+import sttp.model.Uri.QuerySegment
+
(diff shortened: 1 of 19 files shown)


dex/src/main/scala/com/wavesplatform/dex/tool/connectors/AuthServiceRestConnector.scala
```diff
@@ -0,0 +1,55 @@
...
+      "username"   -> jwtPayload.publicKey.base58,
+      "password"   -> s"${jwtPayload.firstTokenExpirationInSeconds}:${jwtPayload.signature}",
+      "scope"      -> jwtPayload.scope.head,
+      "client_id"  -> jwtPayload.clientId
+    ).map { case (k, v) => QuerySegment.KeyValue(k, v) }
+  }
+
+  def getAuthCredentials(maybeSeed: Option[String]): ErrorOr[AuthCredentials] = {
+
+    val keyPair       = KeyPair(crypto secureHash (maybeSeed getOrElse s"minion${ThreadLocalRandom.current.nextInt}" getBytes StandardCharsets.UTF_8))
```
vsuharnikov, line 40: Minor note:

1. `getAuthCredentials` is used only in `checkWsAccountUpdates`
2. We use `maybeSeed` here to get an account or generate a random account. That's all
3. `maybeSeed: Option[String]` in a function doesn't tell us much about "What happens if we pass None"

Suggest to use `seed: String` and generate a random account in `checkWsAccountUpdates`.

PR #285, opened 2020-05-28
DEX-775 Starting issue when there is matching rules

vsuharnikov:


dex/src/test/scala/com/wavesplatform/dex/market/OrderBookActorSpecification.scala
```diff
@@ -109,6 +109,13 @@ class OrderBookActorSpecification
       tp.expectMsg(OrderBookRecovered(pair, Some(50)))
     }
 
+    "recovery - if there is a matching rule - DEX-775" in obcTestWithPrepare(
```
DenisSerebryansky, line 112: DEX-755 in test name :expressionless: 
But in the other hand its ok for a link to the problem description

PR #299, opened 2020-06-19
DEX-776 Checker improvements

DenisSerebryansky:
* Time between blocks calculated automatically and used in repeatRequestOptions
  * DEX REST API optional parameter introduced. It affects DEX WS API uri
  * Node REST API parameter is optional now
  * Used seed for address stream test printed
  * Checker doesn't check swagger pages

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
@@ -18,7 +18,6 @@ import com.wavesplatform.dex.tool.Checker
 import com.wavesplatform.dex.tool.connectors.SuperConnector
 import scopt.{OParser, RenderingMode}
 
-import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
 
 object WavesDexCli extends ScoptImplicits {
@@ -97,11 +96,15 @@ object WavesDexCli extends ScoptImplicits {
           .action((_, s) => s.copy(command = Command.CheckServer.some))
           .text(s"Checks DEX state")
           .children(
+            opt[String]("dex-rest-api")
+              .abbr("dra")
+              .text("DEX REST API uri. Format: scheme://host:port (default scheme will be picked if none was specified)")
+              .valueName("<raw-string>")
+              .action((x, s) => s.copy(dexRestApi = x)),
             opt[String]("node-rest-api")
               .abbr("nra")
               .text("Waves Node REST API uri. Format: scheme://host:port (default scheme will be picked if none was specified)")
               .valueName("<raw-string>")
-              .required()
               .action((x, s) => s.copy(nodeRestApi = x)),
             opt[String]("version")
               .abbr("ve")
@@ -124,12 +127,7 @@ object WavesDexCli extends ScoptImplicits {
               .abbr("as")
               .text("Seed for checking account updates")
               .valueName("<raw-string>")
-              .action((x, s) => s.copy(accountSeed = x.some)),
-            opt[FiniteDuration]("time-between-blocks")
-              .abbr("tbb")
-              .text("Time between blocks in the network, 1 minute will be picked if none was specified")
-              .valueName("<finite duration>")
-              .action((x, s) => s.copy(timeBetweenBlocks = x))
+              .action((x, s) => s.copy(accountSeed = x.some))
           )
       )
     }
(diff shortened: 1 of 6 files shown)


dex/src/main/scala/com/wavesplatform/dex/tool/connectors/AuthServiceRestConnector.scala
```diff
@@ -38,19 +30,21 @@ case class AuthServiceRestConnector(target: String, chainId: Byte) extends RestC
 
   def getAuthCredentials(maybeSeed: Option[String]): ErrorOr[AuthCredentials] = {
 
-    val keyPair       = KeyPair(crypto secureHash (maybeSeed getOrElse s"minion${ThreadLocalRandom.current.nextInt}" getBytes StandardCharsets.UTF_8))
+    val seed          = maybeSeed getOrElse s"minion${ThreadLocalRandom.current.nextInt}"
```
vsuharnikov, line 33: From the task: "Print the account seed. If it wasn't specified, **generate, print** and use it".
So this should be in WavesDexCli.

dex/src/main/scala/com/wavesplatform/dex/tool/connectors/NodeRestConnector.scala
```diff
@@ -13,9 +14,22 @@ import sttp.model.MediaType
...
+  override implicit val repeatRequestOptions: RepeatRequestOptions = {
+    val blocksCount = 5
+    (
+      for {
+        currentHeight <- getCurrentHeight
+        blocksTs <- getBlockHeadersAtHeightRange((currentHeight - blocksCount).max(0), currentHeight)
+          .map(_.map(json => (json \ "timestamp").as[Long]))
+          .ensure("0 or 1 blocks have been forged at the moment, try again later")(_.size > 1)
+      } yield {
+        val timeBetweenBlocks = (blocksTs.zip(blocksTs.tail).map { case (older, newer) => newer - older }.sum / blocksCount * 1.5 / 1000).toInt
```
vsuharnikov, line 28: Or simpler:
```scala
val timeBetweenBlocks = ((blocksTs.last - blocksTs.head) / blocksCount * 1.5 / 1000).toInt
```

dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
```diff
@@ -226,15 +224,15 @@ object WavesDexCli extends ScoptImplicits {
                   _ <- cli.log(
                     s"""
                       |Passed arguments:
+                      |  DEX REST API          : ${args.dexRestApi}
```
vsuharnikov, line 227: If it is empty, we doesn't know, where we are connecting to. Consider to determine the final `dexRestApi` here.

dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
```diff
@@ -226,15 +224,15 @@ object WavesDexCli extends ScoptImplicits {
...
                       |  Waves Node REST API   : ${args.nodeRestApi}
```
vsuharnikov, line 228: This too.

dex/src/main/scala/com/wavesplatform/dex/tool/connectors/AuthServiceRestConnector.scala
DenisSerebryansky: [Generated](https://github.com/wavesplatform/matcher/pull/299/files#diff-6508a5494d9e6d57de82758561b8cc46R33), [printed](https://github.com/wavesplatform/matcher/pull/299/files#diff-dd5ce2fc18ee1b058b6aef14af9e1dcbR190) and [used](https://github.com/wavesplatform/matcher/pull/299/files#diff-dd5ce2fc18ee1b058b6aef14af9e1dcbR188)

dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
DenisSerebryansky: It would be printed below after SuperConnector creation

dex/src/main/scala/com/wavesplatform/dex/tool/connectors/NodeRestConnector.scala
DenisSerebryansky: Done

PR #301, opened 2020-06-23
DEX-705 Graceful shutdown of WS API

DenisSerebryansky:
* Api routes are splitted into 3 values
 * MatcherWebSocketRoute got method gracefulShutdown which sends error messages to all WS connections and then closes them
 * WS API graceful shutdown was embedded into existed shutdown process

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/api/ws/routes/MatcherWebSocketRoute.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/api/ws/routes/MatcherWebSocketRoute.scala
@@ -1,6 +1,7 @@
 package com.wavesplatform.dex.api.ws.routes
 
 import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
 
 import akka.actor.typed.scaladsl.adapter._
 import akka.actor.{ActorRef, typed}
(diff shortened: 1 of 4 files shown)


dex/src/main/scala/com/wavesplatform/dex/api/ws/routes/MatcherWebSocketRoute.scala
```diff
@@ -100,11 +107,25 @@ case class MatcherWebSocketRoute(addressDirectory: ActorRef,
...
-    handleWebSocketMessages { Flow.fromSinkAndSourceCoupled(serverSink, clientSource) }
+    val flow: Flow[Message, TextMessage.Strict, NotUsed] = Flow.fromSinkAndSourceCoupled(serverSink, clientSource)
+    flow.watchTermination()((_, future) => completeWsHandlerTerminationPromise(webSocketHandlerRef, future))
+    handleWebSocketMessages(flow)
   }
 
+  private def completeWsHandlerTerminationPromise(wsHandler: typed.ActorRef[WsHandlerActor.Message], future: Future[Done]): Unit =
+    wsHandlers.computeIfPresent(
+      wsHandler, { (_, p) =>
+        future.onComplete(p.tryComplete)
```
vsuharnikov, line 124: or simpler: `p.completeWith(future)`

PR #314, opened 2020-07-07
DEX-690 Checking tradable balance after creating accounts

ngadiyak:


ngadiyak:
waited for fix in DEX-828

PR #319, opened 2020-07-15
DEX-821 Parallel placing

ngadiyak:


dex-load/src/main/scala/com/wavesplatform/dex/load/utils/package.scala
```diff
@@ -73,13 +73,13 @@ package object utils {
...
-                           amount,
-                           System.currentTimeMillis + 60 * 60 * 24 * 20 * 1000,
-                           settings.defaults.matcherFee)
+      settings.matcherPublicKey,
+      orderType,
+      pair,
+      price,
+      amount,
+      System.currentTimeMillis + 60 * 60 * 24 * 20 * 1000,
+      settings.defaults.matcherFee)
```
vsuharnikov, line 82: Please avoid unnecessary formatting changes. It leads to conflicts.

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -261,34 +283,37 @@ object TankGenerator {
...
-          settings.defaults.matcherFee,
-          null,
-          false
-        )
-        if (i % 50 == 0) println(s"$i orders of $requestsCount have been placed")
-      } catch {
-        case e: Exception => println(e.getMessage)
+    val futures = (0 to requestsCount).map(_ => {
+      Future {
+        blocking {
```
vsuharnikov, line 291: `blocking` makes things worse in this case: it leads to additional thread creation, so the server will face more than 5 simultaneous requests and probably will fail to process some of requests. Also see our chat at 11:16 AM (yesterday).

Suggest to remove `blocking` wrapper.

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -248,7 +268,9 @@ object TankGenerator {
     }
 
     val all = accounts.flatMap(a => {
-      pairs.map(p => { mkTradableBalance(a, p) })
+      pairs.map(p => {
+        mkTradableBalance(a, p)
+      })
```
vsuharnikov, line 273: The compact way :)
```scala
pairs.map(mkTradableBalance(a, _))
```

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -100,7 +112,15 @@ object TankGenerator {
   private def mkPairsAndDistribute(accounts: List[PrivateKeyAccount], pairsFile: Option[File], distributed: Boolean = false): List[AssetPair] = {
     val assets =
       if (Files.notExists(pairsFile.get.toPath)) mkAssets()
-      else readAssetPairs(pairsFile).map(p => { s"${p.getAmountAsset}-${p.getPriceAsset}" }).mkString("-").split("-").toSet.toList
+      else
+        readAssetPairs(pairsFile)
+          .map(p => {
+            s"${p.getAmountAsset}-${p.getPriceAsset}"
+          })
```
vsuharnikov, line 119: Small note about style. You can write
```
.map { p =>
  s"${p.getAmountAsset}-${p.getPriceAsset}"
}
```
which requires less symbols.

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -25,7 +31,12 @@ object TankGenerator {
   private def mkAssets(count: Int = settings.assets.count): List[String] = {
     println(s"Generating $count assets... ")
     val assets = (1 to count).map(_ => mkAsset()).toList
-    waitForHeightArise()
+    val asset  = assets.get(new Random().nextInt(assets.length - 1))
```
vsuharnikov, line 34: If you see the documentation about `nextInt`, you might notice that `nextInt(x)` returns a number in `[0;x)`. So it excludes `x`. The right code here if we need to get the latest asset too:
```
new Random().nextInt(accounts.length)
```

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -77,8 +88,9 @@ object TankGenerator {
       })
     println(s" Done")
 
-    waitForHeightArise()
-    waitForHeightArise()
+    val asset   = assets.get(new Random().nextInt(assets.length - 1))
+    val account = accounts.get(new Random().nextInt(accounts.length - 1))
```
vsuharnikov, line 92: See above

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -11,9 +12,14 @@ import com.wavesplatform.wavesj.matcher.Order.Type
 import play.api.libs.json.{JsError, JsSuccess, JsValue}
 
 import scala.collection.JavaConversions.seqAsJavaList
+import scala.concurrent.duration.DurationInt
+import scala.concurrent._
 import scala.util.Random
 
 object TankGenerator {
+  val threadCount                                                = 5
+  private val executor: ExecutorService                          = Executors.newFixedThreadPool(threadCount)
```
vsuharnikov, line 21: This thread pool creates user threads instead of daemon. So the application won't shutdown if we don't run `executor.shutdown` explicitly.

Consider either create `close` method, or use `ThreadFactory` to create daemon threads.

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: ok

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: ok

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: ok

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: fixed

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -25,7 +31,12 @@ object TankGenerator {
   private def mkAssets(count: Int = settings.assets.count): List[String] = {
     println(s"Generating $count assets... ")
     val assets = (1 to count).map(_ => mkAsset()).toList
-    waitForHeightArise()
+    val asset  = assets.get(new Random().nextInt(assets.length - 1))
+
+    do {
+      waitForHeightArise()
+    } while (services.matcher.getTradableBalance(new AssetPair(asset, "WAVES"), issuer.getAddress()).getOrDefault(asset, 0L) <= 0)
```
DenisSerebryansky, line 38: You will create AssetPair every time you enter the loop
So, extract asset pair

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -77,8 +88,9 @@ object TankGenerator {
       })
     println(s" Done")
 
-    waitForHeightArise()
-    waitForHeightArise()
+    val asset   = assets.get(new Random().nextInt(assets.length - 1))
+    val account = accounts.get(new Random().nextInt(accounts.length - 1))
+    while (services.matcher.getTradableBalance(new AssetPair(asset, "WAVES"), account.getAddress()).getOrDefault(asset, 0L) > 0) waitForHeightArise()
```
DenisSerebryansky, line 93: Here too

ngadiyak: ok

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: ok

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: added shutdown to finally of the main method

PR #344, opened 2020-08-19
DEX-865 Make allure reports nicer

ngadiyak:


vsuharnikov (changes requested):
There are other modules also, including waves-integration-it and other. Please add Story annotation to them too

PR #353, opened 2020-09-01
DEX-842 Coordinated Shutdown

vsuharnikov:
* Moved all shutdown processes to coordinated hooks;
* Added NTP shutdown.

### DIFF
--- a/dex/src/main/scala/com/wavesplatform/dex/queue/KafkaMatcherQueue.scala
+++ b/dex/src/main/scala/com/wavesplatform/dex/queue/KafkaMatcherQueue.scala
@@ -21,8 +21,9 @@ import org.apache.kafka.common.errors.{WakeupException, TimeoutException => Kafk
 import org.apache.kafka.common.serialization._
 
 import scala.concurrent.duration.FiniteDuration
-import scala.concurrent.{Await, ExecutionContext, Future, Promise}
+import scala.concurrent.{ExecutionContext, Future, Promise, blocking}
 import scala.jdk.CollectionConverters._
+import scala.jdk.DurationConverters.ScalaDurationOps
 
 class KafkaMatcherQueue(settings: Settings) extends MatcherQueue with ScorexLogging {
   private val producerThreadPool =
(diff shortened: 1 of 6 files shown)


dex/src/main/scala/com/wavesplatform/dex/queue/KafkaMatcherQueue.scala
```diff
@@ -110,17 +111,17 @@ class KafkaMatcherQueue(settings: Settings) extends MatcherQueue with ScorexLogg
...
-    consumerThreadPool.shutdown()
+    Future {
+      blocking {
+        producer.close(timeout)
+        producerThreadPool.shutdown()
+        consumer.close(timeout.toJava)
+        consumerTask.cancel()
+        consumerThreadPool.shutdown()
+      }
+    }(scala.concurrent.ExecutionContext.global)
```
DenisSerebryansky, line 124: Why global ec?

dex/src/main/scala/com/wavesplatform/dex/queue/LocalMatcherQueue.scala
```diff
@@ -20,7 +20,7 @@ class LocalMatcherQueue(settings: Settings, store: LocalQueueStore, time: Time)
 
   private val executor = Executors.newSingleThreadExecutor {
     new ThreadFactoryBuilder()
-      .setDaemon(true)
+      .setDaemon(false)
```
DenisSerebryansky, line 23: Is this the reason of not working ctrl+c?

DenisSerebryansky (approved):

dex/src/main/scala/com/wavesplatform/dex/queue/KafkaMatcherQueue.scala
vsuharnikov: Because we can't run a Future in `consumerThreadPool`, it closed in the Future :)

dex/src/main/scala/com/wavesplatform/dex/queue/LocalMatcherQueue.scala
vsuharnikov: Nope, actually, I didn't tried reproduce this problem before moving to coordinated shutdown

PR #354, opened 2020-09-02
DEX-835 Node balances stream

DenisSerebryansky:
Notable changes:
* A new version of Matcher is not compatible with the previous! It was unreal to save compatibility;
* Matcher requires two extensions on NODE: its own and balance updates;
* Matcher collects balances from the stream on the server side, not on the extension;
* NODE 1.2.14;

Config:
* Added waves.dex.blockchain-updates-grpc to tune connection to Blockchain Updates Extension;

WebSockets:
* Added a 100 millisecond delay before a connection closes to send the reason;

Matcher HTTP API: added "print" API method.
Added CombinedWavesBlockchainClient to fuse Blockchain Updates and Matcher extensions.
SpendableBalancesActor now only routes messages between balance source and address actors. It will be removed in the future. CombinedWavesBlockchainClient stores balances now.

Blockchain Updates Extension support in blockchainupdates packages:
* Added BlockchainUpdatesClient, GrpcBlockchainEventsObserver;
* Added BlockchainUpdatesStreamControl to have an ability to close or restart the strem, request a next event and etc;
* Added Conversions for gRPC to domain conversions;
* Added StatusTransitions, BlockchainStatus, StatusUpdate, WavesNodeEvent, WavesBlock, WavesBranch, WavesFork classes to update balances and fight forks and disconnects;
* PessimisticPortfolios is moved from the Matcher Extension to the Matcher Server and totally reimplemented due to specifics of extensions;

Matcher Extension:
* Added GetUtxEvents stream. The extension sends the whole state of UTX after a connection, then sends UtxAdded and UtxRemoved;
* Added GetCurrentBlockInfo;
* GetNodeAddress is removed, because we can determine the address using grpc-java classes;
* GetRealTimeBalanceChanges is removed, because we receive these changes from the Blockchain Updates extension;
* Added GetBalances to get separated regular balances and leases;
* Classes are moved to the matcherext packages;
* MatcherExtensionGrpcAsyncClient now calls to gRPC methods using the grpc-java function: channel.newCall to have an ability to tune a request and log the server where a response came from;

Build:
* Packaging and creating artifacts changed to support two extensions (ExtensionPackaging, WavesNodeArtifactsPlugin, waves-ext/docker/dockerfile, etc);
* quickCheckRaw now compiles all projects. root/compile didn't work for all aggregated modules;
* sbt 1.4.3;
* Added MatcherIOUtils.decompressTgz to decompress tgz archives.

Integration tests:
* Better logs in containers;
* When a test fails in a test suite, we do not log canceled tests;
* We do not log calls of debug/print;
* Better logging for exceptions;
* Starting, failing and succeeded tests are printed via "print" method instead of container's shell to be more precise;
* Stabilized tests suites: WsPingPongBaseSuite, TradersTestSuite;
* New test suites: AutoCancelOrderTestSuite, BouncingBalancesTestSuite, 
* Added NoStackTraceCancelAfterFailure - use it instead of CancelAfterFailure to see less garbage in the logs;
* Added to NodeApi: unconfirmedTransactionInfo, rollback, print;
* Added a support of two extensions to toxics;
* DexContainer: added a documentation, when to use *api/async*Api.


waves-ext/src/main/scala/com/wavesplatform/dex/grpc/integration/DEXExtension.scala
```diff
@@ -31,8 +32,12 @@ class DEXExtension(context: ExtensionContext) extends Extension with ScorexLoggi
     val host: String = context.settings.config.as[String]("waves.dex.grpc.integration.host")
     val port: Int    = context.settings.config.as[Int]("waves.dex.grpc.integration.port")
 
+    val ignoredExchangeTxSenderPublicKey: Option[String] = Try {
+      context.settings.config.as[String]("waves.utx.ignore-exchange-sender-pk-in-pessimistic-portfolio")
```
vsuharnikov, line 36: This setting should be a part of `waves.dex` at least.

waves-ext/src/main/scala/com/wavesplatform/dex/grpc/integration/services/WavesBlockchainApiGrpcService.scala
```diff
@@ -2,79 +2,135 @@ package com.wavesplatform.dex.grpc.integration.services
...
+
+  private val utxBalanceUpdates: Observable[Map[Address, Set[Asset]]] = context.utxEvents.map {
+    case TxAdded(tx, diff) =>
+      tx match {
+        case et: ExchangeTransaction if ignoredExchangeTxSenderPublicKey.contains(et.sender.toString) => Map.empty[Address, Set[Asset]]
+        case otherTx =>
+          txDiffs.putIfAbsent(otherTx, diff)
+          getAddressesChangedAssets(diff)
+      }
+    case TxRemoved(tx, _) => Option(txDiffs remove tx).map(getAddressesChangedAssets).getOrElse(Map.empty)
```
vsuharnikov, line 107: We can store `getAddressesChangedAssets(diff)` in `txDiffs` instead of `diff`. Then we won't need to recalculate it in `TxRemoved`

dex-it/src/test/scala/com/wavesplatform/it/sync/BouncingBalancesTestSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+      dex1.api.orderStatus(bobOrder).status shouldBe Status.Accepted
+
+      withClue(s"After rollback order should not be cancelled and balances should not be decreased\n") {
+        wavesNode1.api.rollback(heightInitial, returnTransactionsToUtx = false)
+        wavesNode1.api.currentHeight shouldBe heightInitial
+
+        wavesNode1.api.waitForHeight(heightSecondTransfer)
+        dex1.api.orderStatus(bobOrder).status shouldBe Status.Accepted
+
+        wsc.messages shouldBe empty
```
vsuharnikov, line 107: 👍

waves-ext/src/main/scala/com/wavesplatform/dex/grpc/integration/services/WavesBlockchainApiGrpcService.scala
DenisSerebryansky: Done

waves-ext/src/main/scala/com/wavesplatform/dex/grpc/integration/DEXExtension.scala
DenisSerebryansky: Done

vsuharnikov:
Updated a PR text. There are a lot of changes were made in DEX-956 :)

PR #355, opened 2020-09-03
DEX-784 WebSockets connections rebalance

vsuharnikov:
API changes:
* Added GET /ws/v0/connections to get active connections number;
* Added DELETE /ws/v0/connections to close connections by a filter (currently only "oldest").


dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -54,26 +53,25 @@ import scala.util.Success
...
-                           orderDb: OrderDB,
-                           time: Time,
-                           currentOffset: () => QueueEventWithMeta.Offset,
-                           lastOffset: () => Future[QueueEventWithMeta.Offset],
-                           matcherAccountFee: Long,
-                           apiKeyHash: Option[Array[Byte]],
-                           rateCache: RateCache,
-                           validatedAllowedOrderVersions: () => Future[Set[Byte]],
-                           getActualOrderFeeSettings: () => OrderFeeSettings)(implicit mat: Materializer)
+class MatcherApiRoute(assetPairBuilder: AssetPairBuilder,
```
DenisSerebryansky, line 56: Why did case class become simple class?

DenisSerebryansky (approved):

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
vsuharnikov: Because these classes don't need a comparison, thus we don't need generated `equals` and `hashCode`.

PR #361, opened 2020-09-09
DEX-917 WavesJ 1.0.x migration

DenisSerebryansky:
* Version 1.0.1 used


dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -282,40 +325,54 @@ object TankGenerator {
...
-          settings.defaults.minimalOrderAmount,
-          System.currentTimeMillis + 60 * 60 * 24 * 20 * 1000,
-          settings.defaults.matcherFee,
-          null,
-          false
-        )
-      }.recover {
+    val futures = (0 to requestsCount).map { _ =>
+      val account   = accounts(new Random().nextInt(accounts.length))
+      val assetPair = pairs(new Random().nextInt(pairs.length))
```
vsuharnikov, line 345: Minor: It'd better to create Random instance once before `futures` and use it

vsuharnikov (approved):

PR #362, opened 2020-09-09
DEX-913 Executed Amount of Price Asset

DenisSerebryansky:
* Field totalExecutedAmountOfPriceAsset added to FillingInfo, OrderInfo (v6 implementation introduced), HttpOrderBookHistoryItem, WsOrder and WsFullOrder
 * Unit and integration tests fixed

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/OrderHistoryTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/OrderHistoryTestSuite.scala
@@ -270,11 +270,13 @@ class OrderHistoryTestSuite extends MatcherSuiteBase with TableDrivenPropertyChe
   }
 
   "OrderHistory should" - {
-    "correctly save average weighed price" in {
+    "correctly save average weighed price and total executed amount of price asset" in {
       Seq(alice, bob).foreach { dex1.api.cancelAll(_) }
 
-      def assertAvgWeighedPrice(keyPair: KeyPair, avgWeighedPrices: List[Long]): Unit = {
-        dex1.api.orderHistoryByPair(keyPair, wavesUsdPair, Some(false)).map(_.avgWeighedPrice) should matchTo(avgWeighedPrices)
+      def assertAvgWeighedPrice(keyPair: KeyPair, avgWeighedPricesAndPriceAssetAmount: List[(Long, Long)]): Unit = {
+        dex1.api
+          .orderHistoryByPair(keyPair, wavesUsdPair, Some(false))
+          .map(item => item.avgWeighedPrice -> item.executedAmountOfPriceAsset) should matchTo(avgWeighedPricesAndPriceAssetAmount)
       }
 
       // checking market and limit orders because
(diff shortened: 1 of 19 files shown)


dex-it/src/test/scala/com/wavesplatform/it/sync/OrderHistoryTestSuite.scala
```diff
@@ -292,8 +294,8 @@ class OrderHistoryTestSuite extends MatcherSuiteBase with TableDrivenPropertyChe
 
         placeAndAwaitAtNode(mkOrderDP(mozart, wavesUsdPair, SELL, 95.waves, 2.0), isMarketOrder = isMarketOrder)
 
-        assertAvgWeighedPrice(mozart, List(288L))
-        assertAvgWeighedPrice(salieri, List(270L, 290L, 320L))
+        assertAvgWeighedPrice(mozart, List(288L  -> 259.20.usd))
+        assertAvgWeighedPrice(salieri, List(270L -> 135.usd, 290L -> 29.usd, 320L -> 96.usd))
```
vsuharnikov, line 298: Interesting...
```scala
def executedPrice: Long = counter.price
```

So `salieri` spent USD as expected. But, how much USD should receive `mozart`? The same amount that `salieri` spent, isn't it?

vsuharnikov: In other words: `259.20 != 260 = 135+29+96`

dex/src/main/scala/com/wavesplatform/dex/model/OrderInfo.scala
```diff
@@ -192,7 +210,21 @@ object OrderInfo {
...
+      price,
+      matcherFee,
+      matcherFeeAssetId,
+      timestamp,
+      status,
+      assetPair,
+      orderType,
+      avgWeighedPrice,
+      orderVersion,
+      getTotalExecutedAmountOfPriceAsset(status.filledAmount, avgWeighedPrice)
```
vsuharnikov, line 226: This is okay for old versions of OrderInfo. But in this task we need the exact information. Probably, the best way to achieve this is to nominator instead of avgWeighedPrice

vsuharnikov (changes requested):
Also we need additional checks:
1. Please add checks to `RoundingIssuesTestSuite`
2. Also please add a new test when we have an order with fee in price assets

dex/src/main/scala/com/wavesplatform/dex/model/OrderInfo.scala
DenisSerebryansky: OrderInfo v6 uses `avgWeighedPriceNominator / Order.PriceConstant` (`FillingInfo.totalExecutedAmountOfPriceAsset` field)

vsuharnikov: Interesting... Could we lost some coins because 
```
truncate((avgWeighedPriceNominator + x) / Order.PriceConstant)
!=
truncate(avgWeighedPriceNominator / Order.PriceConstant) + truncate(x / Order.PriceConstant)
```
?

DenisSerebryansky: Yes we could. So I suggets to divide in the end, that is to leave everything as it is

vsuharnikov:
Looked at code, seems fine :)

dex/src/main/scala/com/wavesplatform/dex/model/OrderInfo.scala
```diff
@@ -498,7 +508,7 @@ object OrderInfo {
       assetPair = AssetPair(buf.getAssetId, buf.getAssetId),
       orderType = buf.getAcceptedOrderType,
       orderVersion = buf.get(),
-      totalExecutedPriceAssets = buf.getLong()
+      avgWeighedPriceNominator = new BigInteger(buf.getBytes)
```
vsuharnikov, line 510: This should not work. And probably tests fail because of this. `buf.getBytes` contains a length of array and the array itself.

vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/model/OrderInfo.scala
DenisSerebryansky: No, everything will [work correctly](https://github.com/wavesplatform/matcher/blob/master/dex/src/main/scala/com/wavesplatform/dex/codecs/ByteBufferCodecs.scala#L14)

vsuharnikov (approved):

PR #363, opened 2020-09-10
DEX-919 Swagger errors

vsuharnikov:


dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpMarketDataWithMeta.scala
```diff
@@ -8,20 +8,25 @@ case class HttpMarketDataWithMeta(
     @ApiModelProperty(
       value = "Base58 encoded amount asset ID",
       dataType = "string",
-      example = "8LQW8f7P5d5PZM7GtZEBgaqRPGSzS3DfPuiXrURJ4AJS"
+      example = "8LQW8f7P5d5PZM7GtZEBgaqRPGSzS3DfPuiXrURJ4AJS",
+      required = true
```
DenisSerebryansky, line 12: Why? This is response, not request

dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpMarketStatus.scala
```diff
@@ -6,37 +6,18 @@ import com.wavesplatform.dex.model.{LastTrade, LevelAgg}
...
                             @ApiModelProperty(
                               value = "Side (sell or buy)",
                               dataType = "string",
-                              example = "buy",
-                              allowEmptyValue = true
+                              example = "buy"
                             )
                             lastSide: Option[OrderType],
-                            @ApiModelProperty(
-                              allowEmptyValue = true,
```
DenisSerebryansky, line 25: This was in warnings?

DenisSerebryansky (changes requested):

dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpMarketStatus.scala
vsuharnikov: Yep

dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpMarketDataWithMeta.scala
vsuharnikov: https://swagger.io/specification/#schema-object
```
The Schema Object allows the definition of input and output data types
...
For more information about the properties, see JSON Schema Core and JSON Schema Validation
```
https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.15 :
```
An object instance is valid against this keyword if its property
 set contains all elements in this keyword's array value.
```
As I understood, responses are validated too.

DenisSerebryansky (approved):

PR #366, opened 2020-09-14
DEX-904 Rates updates in WS API

DenisSerebryansky:
* Rates snapshot is sent right after connection
 * Rates updates handled

### DIFF
--- a/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsConnectionTestSuite.scala
+++ b/dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsConnectionTestSuite.scala
@@ -1,7 +1,9 @@
 package com.wavesplatform.it.sync.api.ws
 
 import com.typesafe.config.{Config, ConfigFactory}
+import com.wavesplatform.dex.api.ws.connection.WsConnection
 import com.wavesplatform.dex.api.ws.protocol._
+import com.wavesplatform.dex.domain.asset.Asset
 import com.wavesplatform.dex.domain.asset.Asset.Waves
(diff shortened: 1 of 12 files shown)


dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsConnectionTestSuite.scala
```diff
@@ -11,15 +14,81 @@ import scala.concurrent.{Await, Future}
...
   }
 
+  "Matcher should send rates snapshot after initial message and rates updates" in {
+
+    def assertRatesUpdates(wsc: WsConnection)(expectedRatesUpdates: List[(Map[Asset, Double], Long)]): Unit = {
+      wsc
+        .receiveAtLeastN[WsRatesUpdates](expectedRatesUpdates.size)
+        .map(r => r.rates -> r.updateId) should matchTo(expectedRatesUpdates)
+    }
+
```
ngadiyak, line 33: maybe we need to check the cases:
1. rate is already in config (before the ws-connection was made)
2. asset  doesn't have the rate before we make the ws-connection

vsuharnikov (approved):

dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsConnectionTestSuite.scala
DenisSerebryansky: 1. At least the first test _Rates snapshot_ checks this
2. Actually this PR is about the rates changes notification, so there aren't some problems with this kind of assets

PR #375, opened 2020-09-26
DEX-685 Do not send exchange transactions during start

vsuharnikov:
* Also updated scalafmt


dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
```diff
@@ -71,7 +73,53 @@ class AddressActor(owner: Address,
...
+  }
+
+  private val starting: Receive = eventsProcessing orElse failuresProcessing orElse {
+    case AddressDirectoryActor.StartWork =>
+      activeOrders.values.foreach(x => scheduleExpiration(x.order))
+      started = true
+      context.become(working)
+  }
+
+  private val working: Receive = eventsProcessing orElse failuresProcessing orElse {
```
DenisSerebryansky, line 122: Cases below can be grouped in `commandProcessing` metod

DenisSerebryansky (approved):

dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
vsuharnikov: Will do this next time

PR #378, opened 2020-09-28
DEX-932 Rewrite starting

vsuharnikov:
* The Matcher class was merged with the Application;
* Topological sorting in initialization to get rid of lazy vals;
* ExchangeTransactionCreator.hasMatcherAccountScript by name, because it is loaded after a ETC's instance creation;
* Watching balances after start.


dex/src/main/scala/com/wavesplatform/dex/model/ValidationStages.scala
```diff
@@ -0,0 +1,100 @@
...
+    } yield o
+  }
+
+  def mkSecond(
+      blockchain: AsyncBlockchain,
+      orderBookAskAdapter: OrderBookAskAdapter
+  )(ao: AcceptedOrder, tradableBalance: Map[Asset, Long])(implicit
+      efc: ErrorFormatterContext,
+      ec: ExecutionContext
+  ): Future[Either[MatcherError, Unit]] = {
```
DenisSerebryansky, line 86: FutureResult from `package object effect`?

DenisSerebryansky (approved):

dex/src/main/scala/com/wavesplatform/dex/model/ValidationStages.scala
vsuharnikov: Actually, Future[Either] != EitherT[Future]

PR #382, opened 2020-09-30
DEX-931 Benchmark for processing events during Matcher start

DenisSerebryansky:
* Benchmark added

### DIFF
--- a/dex-jmh/src/test/scala/com/wavesplatform/dex/model/address/AddressActorStartingBenchmark.scala
+++ b/dex-jmh/src/test/scala/com/wavesplatform/dex/model/address/AddressActorStartingBenchmark.scala
@@ -1,0 +1,177 @@
+package com.wavesplatform.dex.model.address
+
+import java.util.concurrent.TimeUnit
+
+import akka.actor.{ActorRef, ActorSystem}
+import akka.pattern.ask
+import cats.syntax.either._
+import com.wavesplatform.dex.actors.address.AddressActor
+import com.wavesplatform.dex.actors.address.AddressActor.Query
+import com.wavesplatform.dex.db.EmptyOrderDB
+import com.wavesplatform.dex.domain.account.KeyPair
+import com.wavesplatform.dex.domain.order.OrderOps._
+import com.wavesplatform.dex.domain.order.{Order, OrderType}
+import com.wavesplatform.dex.error.ErrorFormatterContext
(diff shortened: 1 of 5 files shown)


dex-jmh/src/test/scala/com/wavesplatform/dex/model/address/AddressActorStartingBenchmark.scala
```diff
@@ -0,0 +1,177 @@
...
+      Events.OrderAdded(ao, Events.OrderAddedReason.RequestExecuted, System.currentTimeMillis())
+
+    def getOrderCancelledEvent(ao: AcceptedOrder): Events.OrderCanceled =
+      Events.OrderCanceled(ao, Events.OrderCanceledReason.RequestExecuted, System.currentTimeMillis())
+
+    def getCounter(ao: AcceptedOrder, amount: Long): LimitOrder = {
+      LimitOrder(
+        ao.order
+          .updateType(ao.order.orderType.opposite)
+          .updateSender(clientGen.sample.get)
```
vsuharnikov, line 162: Or create this once above. It doesn't matter, who is the client of the counter order (but it is better to provide a different client)

dex-jmh/src/test/scala/com/wavesplatform/dex/model/address/AddressActorStartingBenchmark.scala
```diff
@@ -0,0 +1,177 @@
...
+              }
+            case OrderEvent.PartiallyFilledAndThenCancelled =>
+              ordersGroup.flatMap { ao =>
+                val firstExecutedEvent = getOrderExecutedEvent(ao, getCounter(ao, ao.amount / 2))
+                val cancelEvent        = getOrderCancelledEvent(firstExecutedEvent.submittedRemaining)
+                Seq(getOrderAddedEvent(ao), firstExecutedEvent, cancelEvent)
+              }
+          }
+        (result :+ eventsGroup) -> newRemaining
+      }
```
vsuharnikov, line 110: I think it is easier to use `Gen.frequency` and create a separate generator for each group, then you won't need a predefined orders, e.g.:
```scala
  val addedGen: Gen[Event] = ???

  Gen.frequency(
    150 -> addedGen,
    /* ... */
  )
```
Also this generator create a shuffled result.

PR #386, opened 2020-10-04
DEX-844 Add private endpoint of getting matcher's config

ngadiyak:


dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+    complete {
+
+      var outputConfig = config.getObject("waves").toConfig
+
+      def pathShouldNotBeDeleted(path: String): Boolean = ("(user|pass|seed|private)".r findFirstIn path).isEmpty
+
+      def modifyConfig(path: String, config: Config): Unit = {
+        val itr = config.root().entrySet().iterator()
+
+        while(itr.hasNext()){
```
vsuharnikov, line 1058: The `while` expression is used in Scala only for a highly optimized code. In this task it could be used for Breadth First Search. But we haven't thousands of keys, so we can use a brute-force solution like this code.

Also iterators aren't used often in Scala, because this code is error prone (e.g. running `next` twice or something else). So it'd better to use `import scala.jdk.CollectionConverters.MapHasAsScala` and `asScala.foreach` instead of `iterator`

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+      }
+
+      modifyConfig("", outputConfig)
+
+      val header = HttpHeader.parse("Content-Type", "application/hocon")  match {
+        case ParsingResult.Ok(header, _) => header
+        case ParsingResult.Error(error)  => throw new Exception(s"Unable to convert to HttpHeader: ${error.summary}")
+      }
+
+      HttpResponse(entity = new ConfigOps(outputConfig).rendered, headers = Seq(header))
```
vsuharnikov, line 1085: Please refer to https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/respond-with-directives/respondWithHeader.html

dex/src/main/resources/application.conf
```diff
@@ -676,3 +676,5 @@ metrics {
     batch-flash-duration = 5s
   }
 }
+
+include "local.conf"
```
vsuharnikov, line 680: Please add a newline

dex/src/test/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRouteSpec.scala
```diff
@@ -277,6 +277,17 @@ class MatcherApiRouteSpec extends RouteSpec("/matcher") with MatcherSpecBase wit
     )
   }
 
+  // getConfig
+  routePath("/debug/config") - {
+    "X-Api-Key is required" in test(
+      { route =>
+        Get(routePath("/debug/config")) ~> route ~> check {
+          status shouldEqual StatusCodes.Forbidden
+        }
```
vsuharnikov, line 286: Check also that it responds with expected `Content-Type`

vsuharnikov (changes requested):
1. Let's move the code that operates with config to `com.wavesplatform.dex.settings.utils.ConfigOps.ConfigOps`:
    ```scala
    /**
     * @note We don't use here BFS, because we don't have to much keys and levels here
     * @param p keyName => Boolean
     */
    def filterKeys(p: String => Boolean): Config
    ```
2. Write unit tests for this code. It should be in dex/src/test/scala/com/wavesplatform/dex/settings/utils/ConfigOpsTestSuite.scala . Use `OrderFeeSettingsCacheSpecification` for template (it one of the simplest test)
3. Filter and render the config in MatcherApiRoute once, because our config can not be changed after start.
4. Please don't forget the formatting

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
vsuharnikov: Also you don't need to create an instance for ConfigOps, just import it and use on `outputConfig`

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+                  modifyConfig(absolutePath, config.getObject(key).toConfig)
+                }
+              }
+            }
+            case _ => if(!pathShouldNotBeDeleted(absolutePath)) outputConfig = outputConfig.withoutPath(absolutePath)
+          }
+        }
+      }
+
+      modifyConfig("", outputConfig)
```
vsuharnikov, line 1078: I think this won't modify `outputConfig`, from the documentation:
```
    /**
     * Clone the object with the given key removed.
     *
     * @param key
     *            key to remove
     * @return a copy of the object minus the specified key
     */
    ConfigObject withoutKey(String key);
```

So this function returns a copy of object. And we don't use this copy

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+    authorizations = Array(new Authorization(SwaggerDocService.apiKeyDefinitionName)),
+    tags = Array("debug"),
+    response = classOf[HttpResponse]
+  )
+  def getConfig: Route = (path("config") & get & withAuth)  {
+    complete {
+
+      var outputConfig = config.getObject("waves").toConfig
+
+      def pathShouldNotBeDeleted(path: String): Boolean = ("(user|pass|seed|private)".r findFirstIn path).isEmpty
```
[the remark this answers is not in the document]
ngadiyak: this method has been removed :) it shorter

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+
+      def modifyConfig(path: String, config: Config): Unit = {
+        val itr = config.root().entrySet().iterator()
```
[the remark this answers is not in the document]
ngadiyak: now i use it once

dex/src/main/resources/application.conf
ngadiyak: ok

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
ngadiyak: changed to `asScala.foreach`

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+
+        while(itr.hasNext()){
+
+          val key = itr.next().getKey
+          val absolutePath = if(path.length > 0) s"$path.$key" else key
```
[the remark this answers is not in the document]
ngadiyak: we filter keys, yes, but Config only has the method "withoutPath"

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+
+          config.root().get(key).valueType() match {
+            case ConfigValueType.OBJECT => {
```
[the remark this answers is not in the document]
ngadiyak: ok

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+              if(pathShouldNotBeDeleted(absolutePath)) {
+                if(config.getObject(key).size() == 0) {
```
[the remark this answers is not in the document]
ngadiyak: ok

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1033,6 +1037,55 @@ class MatcherApiRoute(
...
+                  outputConfig = outputConfig.withoutPath(absolutePath)
+                } else {
+                  modifyConfig(absolutePath, config.getObject(key).toConfig)
+                }
+              }
+            }
+            case _ => if(!pathShouldNotBeDeleted(absolutePath)) outputConfig = outputConfig.withoutPath(absolutePath)
```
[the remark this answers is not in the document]
ngadiyak: ok

dex/src/main/scala/com/wavesplatform/dex/settings/utils/ConfigOps.scala
```diff
@@ -32,6 +33,31 @@ object ConfigOps {
       }
     }
 
+    def filterKeys(part: String): Config = {
```
vsuharnikov, line 36: `part` should be a predicate: `p: String => Boolean`

dex/src/main/scala/com/wavesplatform/dex/settings/utils/ConfigOps.scala
```diff
@@ -32,6 +33,31 @@ object ConfigOps {
...
+      var oc = config
```
vsuharnikov, line 37: Why do you do a copy of `config`?

dex/src/main/scala/com/wavesplatform/dex/settings/utils/ConfigOps.scala
```diff
@@ -32,6 +33,31 @@ object ConfigOps {
...
+          val p = if (b.nonEmpty) s"$b.$k" else k
+
+          if (k.contains(part)) oc = oc.withoutPath(p)
+          else if (v.valueType().equals(ConfigValueType.OBJECT)) {
+            c.getObject(k).isEmpty match {
+              case true => oc = oc.withoutPath(p)
+              case _ => filter(c.getObject(k).toConfig, p)
+            }
+          }
+        })
```
vsuharnikov, line 53: Let's preserve the project's code style:
```scala
c.root().asScala.foreach { obj =>
}
```
It contains less brackets :)

dex/src/main/scala/com/wavesplatform/dex/settings/utils/ConfigOps.scala
```diff
@@ -32,6 +33,31 @@ object ConfigOps {
...
+
+      def filter(c: Config, b: String = ""): Unit = {
+
+        c.root().asScala.foreach(obj => {
+          val k = obj._1
+          val v = obj._2
+          val p = if (b.nonEmpty) s"$b.$k" else k
+
+          if (k.contains(part)) oc = oc.withoutPath(p)
+          else if (v.valueType().equals(ConfigValueType.OBJECT)) {
```
vsuharnikov, line 47: Why `equals` instead of `==` ?

dex/src/main/scala/com/wavesplatform/dex/settings/utils/ConfigOps.scala
```diff
@@ -32,6 +33,31 @@ object ConfigOps {
...
     }
 
+    def filterKeys(part: String): Config = {
+      var oc = config
+
+      def filter(c: Config, b: String = ""): Unit = {
+
+        c.root().asScala.foreach(obj => {
+          val k = obj._1
+          val v = obj._2
```
vsuharnikov, line 43: You can write:
```scala
c.root().asScala.foreach { case (k, v) =>
    // ...
}
```

dex/src/main/scala/com/wavesplatform/dex/settings/utils/ConfigOps.scala
```diff
@@ -32,6 +33,31 @@ object ConfigOps {
...
+          val p = if (b.nonEmpty) s"$b.$k" else k
+
+          if (k.contains(part)) oc = oc.withoutPath(p)
+          else if (v.valueType().equals(ConfigValueType.OBJECT)) {
+            c.getObject(k).isEmpty match {
+              case true => oc = oc.withoutPath(p)
+              case _ => filter(c.getObject(k).toConfig, p)
```
vsuharnikov, line 50: There is only two options. If is enough :)

dex/src/main/scala/com/wavesplatform/dex/settings/utils/ConfigOps.scala
```diff
@@ -32,6 +33,31 @@ object ConfigOps {
       }
     }
 
+    def filterKeys(part: String): Config = {
+      var oc = config
+
+      def filter(c: Config, b: String = ""): Unit = {
```
vsuharnikov, line 39: `filter` should return `Config`, then you don't need to `var`, e.g.:
```scala
def filter(c: Config, b: String = ""): Config = {
 // ...
 oc.withoutPath(p) // returns Config
}
```

vsuharnikov, line 39: `filter` should return `Config`, then you don't need to `var`, e.g.:
```scala
def filter(c: Config, b: String = ""): Config = {
 // ...
 oc.withoutPath(p) // returns Config
}
```

Also you have to use `asScala.foldLeft(c) { ... }` with this approach.

dex-it/src/test/scala/com/wavesplatform/it/sync/api/GetSettingsUtilsTestSuite.scala
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+  override protected def beforeAll(): Unit = {
+    wavesNode1.start()
+    broadcastAndAwait(IssueUsdTx, IssueBtcTx)
+    dex1.start()
+  }
+
+  "GET matcher/debug/config should return config without restricted key-parts (e.g. seed)" in {
+    dex1.api.config.toString should not contain "seed"
```
vsuharnikov, line 24: Does `toString` prints the content of `Config` and not something like `Config@4b32032` ?

vsuharnikov (approved):

PR #412, opened 2020-11-25
DEX-986 Fix status code on error while deleting order book

ngadiyak:


dex/src/main/scala/com/wavesplatform/dex/api/http/entities/MatcherResponse.scala
```diff
@@ -85,7 +85,7 @@ case class OrderCancelRejected(error: MatcherError) extends MatcherResponse(C.Ba
 case object InvalidSignature extends MatcherResponse(C.BadRequest, error.RequestInvalidSignature)
 case class NotImplemented(error: MatcherError) extends MatcherResponse(C.NotImplemented, error)
 case class CanNotPersist(error: MatcherError) extends MatcherResponse(C.ServiceUnavailable, error)
-case class OrderBookUnavailable(error: MatcherError) extends MatcherResponse(C.ServiceUnavailable, error)
+case class OrderBookUnavailable(error: MatcherError) extends MatcherResponse(C.BadRequest, error)
```
vsuharnikov, line 88: `OrderBookUnavailable` is used not only when we are trying to delete an order book, e.g. it was stopped by internal reasons and we can't start it. In these cases `ServiceUnavailable` is appropriate. See usages of `OrderBookUnavailable`.

So I think we need:
1. A new class, or just use `SimpleErrorResponse` with `BadRequest`
2. Use it only for order book deletion (see `MatcherActor`)

dex/src/main/scala/com/wavesplatform/dex/error/MatcherError.scala
```diff
@@ -84,7 +84,7 @@ case class OrderBookBroken(theAssetPair: AssetPair)
       orderBook,
       commonEntity,
       broken,
-      e"The order book for ${Symbol("assetPair") -> theAssetPair} is unavailable, please contact with the administrator"
+      e"The order book for ${Symbol("assetPair") -> theAssetPair} has been deleted or does not exist"
```
vsuharnikov, line 87: Here too. We need a new class, e.g. `OrderBookNotFound` with `orderBook, commonEntity, notFound` trio

vsuharnikov (changes requested):

PR #413, opened 2020-11-25
DEX-973 Order history route tests

ngadiyak:


dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpSuccessfulDeleteHistory.scala
```diff
@@ -0,0 +1,48 @@
...
+import com.wavesplatform.dex.domain.order.Order
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+import play.api.libs.json.{Json, OFormat}
+
+
+@ApiModel(
+  description = "Successful deletion message. Can be HttpSuccessfulDeleteHistory",
+  subTypes = Array(
+    classOf[HttpSuccessfulSingleCancel],
+    classOf[HttpSuccessfulBatchCancel]
```
vsuharnikov, line 12: It looks like the `subTypes` are incorrect. Do they?

dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpSuccessfulDeleteHistory.scala
```diff
@@ -0,0 +1,48 @@
...
+  )
+)
+class HttpSuccessfulDelete {
```
vsuharnikov, line 15: If we have such class, we could use it in Swagger docs? Now I see:
```scala
response = classOf[HttpMessage]
```
for `def deleteHistory`

vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpSuccessfulDeleteHistory.scala
ngadiyak: incorrect, yes

dex/src/main/scala/com/wavesplatform/dex/api/http/entities/HttpSuccessfulDeleteHistory.scala
ngadiyak: ok

vsuharnikov (approved):

PR #422, opened 2020-12-09
DEX-1026 SC6

ngadiyak:


dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -333,6 +336,59 @@ object TankGenerator {
...
+    def mkMassTransfer(transfers: List[Transfer], asset: AssetId): MassTransferTransaction =
+      MassTransferTransaction
+        .builder(transfers.asJava)
+        .assetId(asset)
+        .fee(settings.defaults.massTransferFee + (transfers.size + 1) * settings.defaults.massTransferMultiplier)
+        .version(1)
+        .timestamp(System.currentTimeMillis() + Random.nextLong(100000))
+        .getSignedWith(issuer)
+
+    assetOwners.map { case (assetOwner, asset) =>
```
vsuharnikov, line 355: `foreach`? we don't use the result of `map`.

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -333,6 +336,59 @@ object TankGenerator {
...
+    val assetOwners = accounts.map(_ -> assets(new Random().nextInt(assets.length)))
+    val allRecipients = assetOwners.map(_._1).toSet
+
+    def mkMassTransfer(transfers: List[Transfer], asset: AssetId): MassTransferTransaction =
+      MassTransferTransaction
+        .builder(transfers.asJava)
+        .assetId(asset)
+        .fee(settings.defaults.massTransferFee + (transfers.size + 1) * settings.defaults.massTransferMultiplier)
+        .version(1)
+        .timestamp(System.currentTimeMillis() + Random.nextLong(100000))
```
vsuharnikov, line 352: There could be some duplicates even we use `Random`. Also this code could be copied somewhere in the future. Let's move the timestamp into function's arguments: 
```scala
def mkMassTransfer(transfers: List[Transfer], asset: AssetId, ts: Long): MassTransferTransaction =
  ...
  .timestamp(ts)
```

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -333,6 +336,59 @@ object TankGenerator {
...
+        )
+      } catch { case e: Exception => println(e) }
+    }
+
+    waitForHeightArise()
+
+    val massTransfers =
+      for (_ <- 0 to requestCount / 100) yield assetOwners.map { case (assetOwner, asset) =>
+        mkMassTransfer((allRecipients - assetOwner).map(recipient => Transfer.to(recipient.address(), 1000000L)).toList, AssetId.as(asset))
+      }
```
vsuharnikov, line 379: To make it slightly easier to read:
```scala
val now = System.currentTimeMillis()
val massTransfers =  for {
  i <- 0 to requestCount / 100
  (assetOwner, asset) <- assetOwners
} yield mkMassTransfer(
  transfers = (allRecipients - assetOwner).map(recipient => Transfer.to(recipient.address(), 1000000L)).toList,
  asset = AssetId.as(asset),
  ts = now + i
)
```

dex-load/src/main/scala/com/wavesplatform/dex/load/WavesDexLoadCli.scala
```diff
@@ -179,90 +183,123 @@ object WavesDexLoadCli extends ScoptImplicits {
...
-                val clients: Seq[WsCollectChangesClient] = cli.wrapByLogs[Id, Seq[WsCollectChangesClient]]("Creating clients.. ") {
-                  WsAccumulateChanges.createClients(args.dexWsApi, args.feederFile, args.accountsNumber)
+                def checkLeaps() = {
+                  val clients: Seq[WsCollectChangesClient] =
+                    WsAccumulateChanges.createClients(args.dexWsApi, args.feederFile, args.accountsNumber, args.wsCheckType)
+
+                  val withLeaps =
+                    try {
+                      Future.traverse(clients)(_.run())
+                      Await.result(GlobalTimer.instance.sleep(args.collectTime), Duration.Inf)
```
vsuharnikov, line 193: We can use `Thread.sleep` here. `GlobalTimer.instance.sleep` is used in asynchronous code with `Futures` :)

dex-load/src/main/scala/com/wavesplatform/dex/load/ws/WsCollectChangesClient.scala
```diff
@@ -19,16 +19,17 @@ class WsCollectChangesClient(apiUri: String, address: String, aus: String, obs:
...
   private val orderBookUpdates = mutable.AnyRefMap.empty[AssetPair, WsOrderBookChanges]
+  val addressUpdateLeaps = scala.collection.mutable.ArrayBuffer.empty[String]
   @volatile private var gotPings = 0
 
   private val receive: Function[WsServerMessage, Option[WsClientMessage]] = {
     case x: WsPingOrPong => gotPings += 1; x.some
     case x: WsInitial => log.info(s"Connection id: ${x.connectionId}"); none
     case x: WsError => log.error(s"Got error: $x"); throw new RuntimeException(s"Got $x")
-    case diff: WsAddressChanges => accountUpdates = merge(accountUpdates, diff); none
+    case diff: WsAddressChanges => collectLeaps(accountUpdates, diff); accountUpdates = merge(accountUpdates, diff); none
```
vsuharnikov, line 32: Could we collect also the total number of `WsAddressChanges` to get the ratio of leaps?

dex-load/src/main/scala/com/wavesplatform/dex/load/WavesDexLoadCli.scala
```diff
@@ -312,7 +349,8 @@ object WavesDexLoadCli extends ScoptImplicits {
     requestsFile: File = new File(s"requests-${System.currentTimeMillis}.txt"),
     dexRestApi: String = "",
     collectTime: FiniteDuration = 5.seconds,
-    wsResponseWaitTime: FiniteDuration = 5.seconds
+    wsResponseWaitTime: FiniteDuration = 5.seconds,
+    wsCheckType: Int = 1
```
vsuharnikov, line 353: 1. Numbers are not descriptive. We use sealed traits ([enums are bad](https://medium.com/@yuriigorbylov/scala-enumerations-hell-5bdba2c1216)) to describe different type of elements. See `WavesDexCli.SeedFormat` for example.
2. You changed the default behavior of Check to `checkLeaps`. This is not a good practice when you introduce new features, because it can break places where it is used. Could this break other tests? Even if not, I suggest to not change the default behavior and update the command in SC6 runner with specifying the appropriate `ws-check-type`.

vsuharnikov (changes requested):

dex-load/src/main/scala/com/wavesplatform/dex/load/GatlingFeeder.scala
```diff
@@ -52,14 +52,16 @@ object GatlingFeeder {
...
+      val source = Source.fromFile(pairsFile)
       val pairs = Random.shuffle(source.getLines().toVector)
+      source.close()
       require(numberPerClient <= pairs.size, "numberPerClient > available asset pairs in file")
       pairs.take(numberPerClient).map(x => s"""{"T":"obs","S":"$x","d":100}""").mkString(";")
-    } finally source.close()
-  }
+    } catch {
+      case _: Throwable => ""
+    }
```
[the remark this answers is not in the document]
ngadiyak: fixed

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: ok!

dex-load/src/main/scala/com/wavesplatform/dex/load/ws/WsCollectChangesClient.scala
ngadiyak: fixed

dex-load/src/main/scala/com/wavesplatform/dex/load/WavesDexLoadCli.scala
ngadiyak: ok

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
ngadiyak: https://github.com/wavesplatform/matcher/blob/DEX-1026-sc6/dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala#L373

vsuharnikov (approved):

vsuharnikov (approved):

PR #436, opened 2021-01-12
DEX-996 Refactoring raw-api tests

ngadiyak:


dex-it/src/test/scala/com/wavesplatform/it/matcher/api/http/HttpApiSuiteBase.scala
```diff
@@ -0,0 +1,89 @@
...
+      mkOrder(alice, wavesUsdPair, BUY, 10.waves, 3.usd)
+    ).map { o =>
+      placeAndAwaitAtDex(o)
+      o.idStr()
+    }
+
+  protected def shouldReturnErrorWithoutApiKeyHeader(): Unit =
+    "should return an error without X-API-KEY" in {
+
+      val r = this.getClass.getSimpleName match {
```
vsuharnikov, line 24: This code:
```scala
this.getClass.getSimpleName match {
```
looks bad, because you can forget to add a new class to this pattern matching. Often such intention could be represented through an inheritance.

1. If all descendants of `HttpApiSuiteBase` has `shouldReturnErrorWithoutApiKeyHeader` and `shouldReturnErrorWithIncorrectApiKeyValue `, then you should just write:
    ```scala
    "should return an error without X-API-KEY" in {
      // ...
    ```
    instead of a method
2. Let's create traits with these tests and mix them to the classes, which require these checks
3. This code:
    ```scala
    val r = this.getClass.getSimpleName match {
    ```
    could be replace by an abstract method:
    ```scala
    def requestWithoutApiKeyHeader(): ...
    ```
    then the test will be:
    ```scala
    "should return an error without X-API-KEY" in {
      val r = requestWithoutApiKeyHeader()
      validateAuthorizationError(r)
    }
    ```
    same for `shouldReturnErrorWithIncorrectApiKeyValue`

dex-it/src/test/scala/com/wavesplatform/it/matcher/api/http/HttpApiSuiteBase.scala
```diff
@@ -0,0 +1,89 @@
...
+import com.wavesplatform.it.MatcherSuiteBase
+
+trait HttpApiSuiteBase extends MatcherSuiteBase with RawHttpChecks {
+
+  protected def placeAndGetIds(): Set[String] =
+    Set(
+      mkOrder(alice, wavesUsdPair, BUY, 10.waves, 1.usd),
+      mkOrder(alice, wavesUsdPair, BUY, 10.waves, 2.usd),
+      mkOrder(alice, wavesUsdPair, BUY, 10.waves, 3.usd)
+    ).map { o =>
```
vsuharnikov, line 16: How about
```scala
(1 to 3).map(i => mkOrder(alice, wavesUsdPair, BUY, 10.waves, i.usd))
```
?

vsuharnikov (changes requested):

dex-it/src/test/scala/com/wavesplatform/it/matcher/api/http/HttpApiSuiteBase.scala
ngadiyak: ok

vsuharnikov (approved):

PR #455, opened 2021-02-03
DEX-1066 Use kanela-agent instead of aspectj

ngadiyak:


dex/build.sbt
```diff
@@ -146,6 +150,7 @@ inConfig(Debian)(
 
       Seq(upstartScript -> s"/etc/init/${packageName.value}.conf").map(packageMapping(_).withConfig().withPerms("644"))
     },
+    javaAgents += "io.kamon" % "kanela-agent" % "1.0.7",
```
vsuharnikov, line 153: Try to remove this

vsuharnikov (changes requested):

PR #460, opened 2021-02-09
DEX-1074 Async calls to AssetPairsDB

vsuharnikov:


vsuharnikov:
The problem is in the unstable test, will fix in a next PR

PR #466, opened 2021-02-15
DEX-1067 Up node version to 1.2.17

ngadiyak:


waves-ext/src/main/scala/com/wavesplatform/transaction/smart/script/ScriptRunnerFixed.scala
```diff
@@ -0,0 +1,136 @@
...
+  def apply(
+    in: TxOrd,
+    blockchain: Blockchain,
+    script: Script,
+    isAssetScript: Boolean,
+    scriptContainerAddress: Environment.Tthis,
+    complexityLimit: Int = Int.MaxValue,
+    default: EVALUATED = TRUE
+  ): (Log[Id], Either[ExecutionError, EVALUATED]) =
+    apply(
```
ngadiyak, line 36: That method is for using in the Node.
We don't use it in the Matcher because we pass denied block-chain as the parameter.

vsuharnikov (approved):

PR #469, opened 2021-02-17
DEX-970 Handle invalid depth parameter at getOrderBook

ngadiyak:


dex/src/main/scala/com/wavesplatform/dex/error/MatcherError.scala
```diff
@@ -595,6 +595,9 @@ case class SubscriptionsLimitReached(limit: Int, id: String)
 case class InvalidAddress(reason: String)
     extends MatcherError(address, commonEntity, commonClass, e"Provided address in not correct, reason: ${Symbol("reason") -> reason}")
 
+case class InvalidDepth(reason: String)
+  extends MatcherError(depth, commonEntity, commonClass, e"Provided depth in not correct, reason: ${Symbol("reason") -> reason}")
```
vsuharnikov, line 599: Hm... I think, it is a part of request (as `RequestArgumentInvalid`). Consider to switch to `(request, depth, commonClass)`

ngadiyak: ok

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -386,9 +387,17 @@ class MatcherApiRoute(
...
-    parameters("depth".as[Int].?) { depth =>
-      withAssetPair(pairOrError, redirectToInverse = true, depth.fold("")(d => s"?depth=$d")) { pair =>
-        complete(orderBookHttpInfo.getHttpView(pair, MatcherModel.Normalized, depth))
+    parameters("depth".as[String].?) { depth =>
+
+      def response(d: Int = 0) = withAssetPair(pairOrError, redirectToInverse = true, s"?depth=$d") { pair =>
+        complete(orderBookHttpInfo.getHttpView(pair, MatcherModel.Normalized, Some(d)))
+      }
+
+      allCatch.opt(depth.get.toInt) match {
```
vsuharnikov, line 396: `?depth=a` won't work here and will be caught in `case None =>`. Also `>= 0` is not correct here, because `depth` should be positive, but this condition is for non-negative values.

The working alternative:
```scala
depth match {
  case None => response()
  case Some(depth) =>
    depth.toIntOption match {
      case None => complete(InvalidDepth(s"Depth value '${depth.take(5)}' must be an Integer"))
      case Some(depth) => 
        if (depth > 0) response(depth)
        else complete(InvalidDepth(s"Depth value '$depth' must be positive"))
    }
}
```

vsuharnikov (changes requested):

vsuharnikov (approved):

PR #485, opened 2021-03-29
Dex 1105 app name to db connection

JennaWestenra:


dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
```diff
@@ -210,6 +212,17 @@ class PostgresHistoryDatabaseTestSuite extends MatcherSuiteBase with HasPostgres
     ctx.run(querySchema[EventRecord]("events").delete)
   }
 
+  "Node should provide its name to postgres" in {
+
+    val buyOrder = mkOrder(bob, wavesBtcPair, BUY, 270477189L, 28259L)
```
ngadiyak, line 217: amount and price values could be written like (e.g. 1.waves, or 10.btc)

dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
```diff
@@ -210,6 +212,17 @@ class PostgresHistoryDatabaseTestSuite extends MatcherSuiteBase with HasPostgres
...
+    placeAndAwaitAtDex(buyOrder)
```
ngadiyak, line 218: can we get app name from connection without placing an order?

dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
```diff
@@ -210,6 +212,17 @@ class PostgresHistoryDatabaseTestSuite extends MatcherSuiteBase with HasPostgres
...
+
+    findUserWithApplicationName().value shouldBe PgStatActivity(customUser, customAppName)
+
+    Seq(alice, bob).foreach(dex1.api.cancelAll(_))
```
ngadiyak, line 222: we have placed only one order, there is no need to use the cycle i guess

dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
JennaWestenra: No. When I ran app, it didn't connect to db at start, only when I tried to place an order from code. Tried to remove order from this test, got the same behavior: test's failed because no such connection was found.

dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
JennaWestenra: Fixed. Also changed it in the next test case

dex-it/src/test/scala/com/wavesplatform/it/sync/PostgresHistoryDatabaseTestSuite.scala
JennaWestenra: Yes, fixed it

ngadiyak (approved):

PR #490, opened 2021-04-06
DEX-1103 Wait until matcher starts checks

alexander-branevskiy:


dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
```diff
@@ -320,6 +320,11 @@ object WavesDexCli extends ScoptImplicits {
               .valueName("<raw-string>")
               .required()
               .action((x, s) => s.copy(configPath = x)),
+            opt[String]("api-key")
```
ngadiyak, line 323: I think the key should be written by user via readSecretFromStdIn

vsuharnikov: Agree. We need to read this secret carefully and prevent it from leaking to the bash history.

We have same parameter in `CreateApiKey`, but it covers another case:
* An admin does this once;
* When an admin should see the input and not make a mistake.

vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
alexander-branevskiy: resolved

vsuharnikov (approved):

JennaWestenra (approved):

PR #496, opened 2021-04-07
DEX-1121 Fix OrdersFromScriptedAccTestSuite

ngadiyak:


dex-it/src/test/scala/com/wavesplatform/it/sync/smartcontracts/OrdersFromScriptedAccTestSuite.scala
```diff
@@ -34,6 +34,8 @@ class OrdersFromScriptedAccTestSuite extends MatcherSuiteBase {
   private val aliceAsset = IssuedAsset(aliceAssetTx.id())
   private val aliceWavesPair = AssetPair(aliceAsset, Waves)
 
+  private val orderV2 = mkOrder(bob, aliceWavesPair, OrderType.BUY, 500, 2.waves * Order.PriceConstant, smartTradeFee, version = 2)
```
vsuharnikov, line 37: `def`, otherwise you create the one order and use it everywhere

ngadiyak: fixed

PR #513, opened 2021-04-15
DEX-1134 Try to fix DexClientFaultToleranceTestSuite

alexander-branevskiy:


project/Dependencies.scala
```diff
@@ -147,6 +148,7 @@ object Dependencies {
   private val sttpClient = sttpClientModule("core")
   private val sttpPlayJson = sttpClientModule("play-json")
   private val sttpAsyncHttpClient = sttpClientModule("async-http-client-backend-future")
+  private val asyncHttpClient = "org.asynchttpclient" % "async-http-client" % Version.asyncHttpClient
```
vsuharnikov, line 151: Why do we need this? This dependency retrieved transitively through `async-http-client-backend-future`, so we need to update `sttpClient` version, e.g. to 3.2.3: https://github.com/softwaremill/sttp/blob/v3.2.3/build.sbt

project/Dependencies.scala
```diff
@@ -46,6 +46,7 @@ object Dependencies {
 
     val sttp = "1.7.2"
```
vsuharnikov, line 47: Also please remove this, it is not used anymore )

vsuharnikov (changes requested):

project/Dependencies.scala
alexander-branevskiy: resolved

project/Dependencies.scala
alexander-branevskiy: resolved

PR #518, opened 2021-04-19
DEX-1120 Bouncing balances bug

vsuharnikov:


waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/clients/matcherext/UtxEventConversions.scala
```diff
@@ -31,4 +39,9 @@ object UtxEventConversions extends ScorexLogging {
         none
     }
 
+  // DEX-1120
+  private def isFalsePositive(e: UtxEvent.Update.Removed.Reason): Boolean =
+    // See WavesBlockchainApiGrpcService.canRetry
+    e.name == "OrderValidationError" && e.message.startsWith("Too much")
```
alexander-branevskiy, line 45: Is "Too much" sufficient here? Can we have collisions with other messages?

alexander-branevskiy (approved):

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/clients/matcherext/UtxEventConversions.scala
vsuharnikov: Yep, there are no other errors with "Too much": https://github.com/wavesplatform/Waves/search?q=OrderValidationError

JennaWestenra (approved):

PR #526, opened 2021-04-22
DEX-1155 Fix sync with node

alexander-branevskiy:


JennaWestenra (approved):

vsuharnikov (changes requested):
I've found, that we can't forcefully restart the synchronization.
This happens because a channel in `GrpcBlockchainUpdatesControlledStream.BlockchainUpdatesObserver.scheduleRestart` isn't stopped in `stopGrpcObserver` even we call `cancel`, see the last lines in `Issue` before `Solution` in the task for more details. 
So `onError` isn't called and `internalSystemStream` don't receive `SystemEvent.Stopped`. This causes `CombinedStream` to do nothing.

Let's close the channel in `GrpcBlockchainUpdatesControlledStream.stopGrpcObserver and stop`. Also you could try to share this code between `stop` and `stopGrpcObserver`.

dex-it/src/test/scala/com/wavesplatform/it/sync/orders/OrderDynamicFeeTestSuite.scala
```diff
@@ -802,7 +802,7 @@ class OrderDynamicFeeTestSuite extends OrderFeeBaseTestSuite {
 
       broadcastAndAwait(mkTransfer(alice, bob, defaultAssetQuantity / 2, eth, 0.005.waves))
 
-      dex1.restartWithNewSuiteConfig(ConfigFactory.parseString("waves.dex.order-fee.-1.mode = percent"))
+      dex1.restartWithNewSuiteConfig(ConfigFactory.parseString("waves.dex.order-fee.-1.mode = percent").withFallback(dexInitialSuiteConfig))
```
vsuharnikov, line 805: 👍

vsuharnikov (approved):

PR #531, opened 2021-04-28
DEX-610-1143 Performance tests refactoring 

ngadiyak:


dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -67,6 +70,12 @@ object TankGenerator {
     accounts.toList
   }
 
+  def waitForEmptyUtx(): Unit = {
+    println("Waiting for empty UTX...")
+    while (node.getUtxSize() > 0)
+      waitForHeightArise()
```
vsuharnikov, line 76: 👍

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala
```diff
@@ -102,7 +108,7 @@ object TankGenerator {
         assets
           .combinations(2)
           .map {
-            case List(aa, pa) => if (pa < aa) (aa, pa) else (pa, aa)
+            case List(aa, pa) => if (IssuedAsset(pa.getBytes()).compatId < IssuedAsset(aa.getBytes()).compatId) (aa, pa) else (pa, aa)
```
vsuharnikov, line 111: 1. Interesting, how it works, I didn't see `AssetPairBuilder.assetIdOrdering` import here.
2. This should not work, because:
```scala
scala> val xs = Array[Byte](1, 2, 3)
     | val ys = java.util.Base64.getEncoder.encodeToString(xs).getBytes
     | java.util.Arrays.equals(ys, xs)
val xs: Array[Byte] = Array(1, 2, 3)
val ys: Array[Byte] = Array(65, 81, 73, 68)
val res1: Boolean = false
```
3. How it should look:
3.1. Use `map` and `Asset.fromString` to parse each asset in `assets`. Throw an exception if it can't decode the asset id. So you will have `List[Asset]`
3.2. Import `AssetPairBuilder.assetIdOrdering` to compare ids.
3.3. Compare ids with `<` as usual
3.4. Use `Asset.toString` to convert `(Asset, Asset)` into `(String, String)`

vsuharnikov (changes requested):
Also, let's use our common naming for branches next time. If you took small fixes from the other task - you could link the task and write something like "Fixed in DEX-XXX"

dex-load/src/main/scala/com/wavesplatform/dex/load/TankGenerator.scala

vsuharnikov (approved):

PR #532, opened 2021-04-28
DEX-1154 Hide kanela banner from logs

ngadiyak:


dex/src/main/resources/application.conf
```diff
@@ -777,6 +777,11 @@ scala.concurrent.context {
   maxExtraThreads = "256"
 }
 
+kanela {
+    show-banner = false
+    log-level = "WARNING"
```
alexander-branevskiy, line 782: why do we need to set log level here?

JennaWestenra (approved):

dex/src/main/resources/application.conf
ngadiyak: i guess we needn't INFO logs

vsuharnikov: https://github.com/kamon-io/kanela/search?q=Logger.info I see nothing useful too
But WARNING is the default logging level: https://github.com/kamon-io/kanela/blob/master/agent/src/main/resources/reference.conf
Let's remove it.

alexander-branevskiy (approved):

vsuharnikov (approved):

PR #535, opened 2021-04-29
DEX-1156 Cancel active orders before execution in checker

ngadiyak:


dex/src/main/scala/com/wavesplatform/dex/tool/Checker.scala
```diff
@@ -165,6 +165,23 @@ class Checker(superConnector: SuperConnector) {
...
+    def checkEmptyListOfActiveOrders(): ErrorOr[Boolean] =
+      for {
+        _ <- dexRest.getActiveOrdersByPair(env.matcherKeyPair, assetPairInfo.assetPair)
+          .map {
+            case left: Seq[JsValue] => left.foreach { e =>
+                val id = (e \ "id").as[ByteStr]
+                dexRest.cancelOrder(id, assetPairInfo.assetPair, env.matcherKeyPair)
+                dexRest.waitForOrderStatus(id, assetPairInfo.assetPair, OrderStatus.Cancelled.name)
+              }
+            case _ => _
```
vsuharnikov, line 177: This shouldn't compile :)

dex/src/main/scala/com/wavesplatform/dex/tool/Checker.scala
```diff
@@ -165,6 +165,23 @@ class Checker(superConnector: SuperConnector) {
     val submitted = mkMatcherOrder(assetPairInfo.assetPair, SELL)
     val submittedId = submitted.id()
 
+    def checkEmptyListOfActiveOrders(): ErrorOr[Boolean] =
+      for {
+        _ <- dexRest.getActiveOrdersByPair(env.matcherKeyPair, assetPairInfo.assetPair)
+          .map {
+            case left: Seq[JsValue] => left.foreach { e =>
```
vsuharnikov, line 172:
```scala
def getActiveOrdersByPair(keyPair: KeyPair, assetPair: AssetPair): ErrorOr[Seq[JsValue]]
type ErrorOr[A] = Either[String, A]
```

So you'll receive either `Left`, or `Right`, but not just `Seq`

dex/src/main/scala/com/wavesplatform/dex/tool/Checker.scala
```diff
@@ -165,6 +165,23 @@ class Checker(superConnector: SuperConnector) {
...
+                val id = (e \ "id").as[ByteStr]
+                dexRest.cancelOrder(id, assetPairInfo.assetPair, env.matcherKeyPair)
+                dexRest.waitForOrderStatus(id, assetPairInfo.assetPair, OrderStatus.Cancelled.name)
+              }
+            case _ => _
+          }
+
+        ids <- dexRest
+          .getActiveOrdersByPair(env.matcherKeyPair, assetPairInfo.assetPair)
+          .ensure("Count of active orders > 0")(_.isEmpty)
```
vsuharnikov, line 182: This check could fail, if Matcher haven't yet processed cancels. We should check this in the loop with:
1. E.g. 5 attempts
2. A delay of 1 second between attempts

vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/tool/Checker.scala
ngadiyak: fixed

dex/src/main/scala/com/wavesplatform/dex/tool/Checker.scala
ngadiyak: fixed

dex/src/main/scala/com/wavesplatform/dex/tool/Checker.scala
ngadiyak: i removed retry, since we wait Cancel status there is no need to retry

vsuharnikov (approved):

PR #544, opened 2021-05-13
DEX-1167 SystemInformationReporter exposes a private information

ngadiyak:


dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
```diff
@@ -1168,7 +1165,7 @@ class MatcherApiRoute(
   )
   def getMatcherConfig: Route = (path("config") & get & withAuth) {
     complete {
-      HttpEntity(filteredConfig.rendered).withContentType(CustomContentTypes.`application/hocon`)
+      HttpEntity(config.rendered).withContentType(CustomContentTypes.`application/hocon`)
```
vsuharnikov, line 1168: Consider rename `config` in the constructor to `safeConfig` to make things clear. Otherwise someone might use `config` and think that it contains all keys.

dex/src/main/scala/com/wavesplatform/dex/logs/SystemInformationReporter.scala
```diff
@@ -46,7 +31,7 @@ object SystemInformationReporter extends ScorexLogging {
     ).map { x =>
       x -> System.getProperty(x)
     } ++ Seq(
-      "Configuration" -> configForLogs.root.render(renderOptions)
+      "Configuration" -> config.root.render(renderOptions)
```
vsuharnikov, line 34: Here too.

dex/src/main/scala/com/wavesplatform/dex/Application.scala
```diff
@@ -563,6 +564,9 @@ object Application {
     val configFile = args.headOption
     val (config, settings) = loadApplicationConfig(configFile.map(new File(_)))
 
+    val excludedConfigKeys = config.getStringList("waves.dex.secure-keys").asScala.toSet
```
vsuharnikov, line 567: To make this code consistent with other, let's:
1. Add the config parameter `secure-keys` to `MatcherSettings`
2. Add a check into `MatcherSettingsSpecification`. Also you have to provide a non-default value for `MatcherSettingsSpecification` in `BaseSettingsSpecification.configWithSettings`

vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/api/http/routes/MatcherApiRoute.scala
ngadiyak: +

dex/src/main/scala/com/wavesplatform/dex/logs/SystemInformationReporter.scala
ngadiyak: ok

dex/src/main/scala/com/wavesplatform/dex/Application.scala
ngadiyak: fixed

vsuharnikov: So, you can use `settings.secureKeys` instead of `getStringList`.

dex/src/test/scala/com/wavesplatform/dex/settings/BaseSettingsSpecification.scala
```diff
@@ -12,6 +12,8 @@ class BaseSettingsSpecification extends AnyFlatSpec {
   def getSettingByConfig(conf: Config): Either[String, MatcherSettings] =
     Try(ConfigSource.fromConfig(conf).at("waves.dex").loadOrThrow[MatcherSettings]).toEither.leftMap(_.getMessage)
 
+  val correctSecureKeys: Set[String] = Set("user", "pass", "seed", "private", "java", "sun", "api")
```
vsuharnikov, line 15: The values is equal to the default one. In these tests we use a different value.

vsuharnikov (changes requested):

vsuharnikov (approved):

PR #549, opened 2021-05-16
DEX-1165 Smoke Tests

ngadiyak:


jenkins/tests-multiple-versions.Jenkinsfile
```diff
@@ -18,15 +18,15 @@ pipeline {
         PATH = "${env.SBT_HOME}/bin:${env.PATH}"
         SCALATEST_INCLUDE_TAGS = 'com.wavesplatform.it.tags.DexMultipleVersions'
         KAFKA_SERVER = "${KAFKA_SERVER}"
-        OTHER_DEX_IMAGE = "${OTHER_DEX_IMAGE}"
-        OTHER_NODE_IMAGE = "${OTHER_NODE_IMAGE}"
+        OTHER_DEX_IMAGE = "${REGISTRY}/waves/dex/${OTHER_DEX_IMAGE}"
+        OTHER_NODE_IMAGE = "${REGISTRY}/waves/dex/${OTHER_NODE_IMAGE}"
```
vsuharnikov, line 22: Do these changes affect `post.always` below in this file and in `tests-smoke.Jenkinsfile` and `tests-specified-versions.Jenkinsfile`?

ngadiyak: we have only echo to allure result in the post section, it's not a problem if in the allure we will have a correct ursl to registry

PR #557, opened 2021-05-24
DEX-1166 Enumerate logs in AddressActor

ngadiyak:


vsuharnikov (changes requested):
Forgot here:
```scala
val newReserved = balances.reserved.filter { case (asset, _) => reservedAssets.contains(asset) }      
log.info(s"[Balance] 💵: ${format(balances.tradableBalance(cumulativeDiff.keySet).xs)}; e: ${format(cumulativeDiff)}, ov: ${format(newReserved)}"      )
```

vsuharnikov (approved):

PR #559, opened 2021-05-24
DEX-1180 Improve DatabaseBackwardCompatTestSuite

JennaWestenra:


dex-it/src/test/scala/com/wavesplatform/it/sync/compat/DatabaseBackwardCompatTestSuite.scala
```diff
@@ -13,38 +14,77 @@ import scala.util.Random
...
-      orders.size shouldBe >(0)
+      orders.size shouldBe >(100) // if its > 100, then there will be not only limit or place orders
     }
 
     orders.groupBy(_.assetPair).valuesIterator.foreach { orders =>
       dex2.api.waitForOrder(orders.last)(_.status != Status.NotFound)
     }
+
+    markup("cancel one order at DEX2")
+    val orderToCancel = orders.head
```
vsuharnikov, line 49: This probably doesn't work, because this order could be matched, thus we don't save the Cancel command to the queue and don't test here.

One approach is to get only active orders of alice and bob, pick a random active and cancel it just by `dex1.api.cancel`

vsuharnikov (changes requested):

dex-it/src/test/scala/com/wavesplatform/it/sync/compat/DatabaseBackwardCompatTestSuite.scala
```diff
@@ -13,38 +14,77 @@ import scala.util.Random
...
+      ts,
+      ts + 6000000,
+      matcherFee,
+      2
+    )
+    executeCommands(Seq(MatcherCommand.Place(dex2, additionalOrder)), timeout = 10.second) // it must be short operation
+    dex2.api.waitForOrder(additionalOrder)(_.status != Status.NotFound)
+
+    markup("delete orderbook for additional asset pair at DEX2")
+    executeCommands(Seq(MatcherCommand.DeleteOrderBook(dex2, additionalAssetPair)), timeout = 10.second) // it must be short operation
```
vsuharnikov, line 71: Also you can delete an order book just by `dex1.api.deleteOrderBook`

vsuharnikov (approved):

PR #561, opened 2021-05-24
DEX-1182 NFT assets in Address stream tests

ngadiyak:


dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsAddressStreamTestSuite.scala
```diff
@@ -322,6 +323,61 @@ class WsAddressStreamTestSuite extends WsSuiteBase with TableDrivenPropertyCheck
...
+        "via IssueTx shouldn't be in the address stream" in {
+          val acc = mkAccountWithBalance(10.waves -> Waves)
+          val IssueResults(txIssue, _, issuedAsset) = mkIssueExtended(acc, "testAssetNT", 1L, 0)
+
+          broadcastAndAwait(txIssue)
+
+          val wsc = mkWsAddressConnection(acc)
+          assertChanges(wsc)(
+            Map(Waves -> WsBalances(9, 0)),
+            Map(issuedAsset -> WsBalances(1, 0))
```
ngadiyak, line 336: should be removed after we had added filter for nft

vsuharnikov (approved):

PR #566, opened 2021-05-31
DEX-1104 Fix Docker ports

JennaWestenra:


dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,81 @@
...
+
+      case None => Left(AvailablePortsNotFound())
+    }
+  }
+
+  private def parsePortRange(): Range.Inclusive = {
+    val limits = stringRange.split('-')
+    if (limits.length != 2) throw new IllegalArgumentException(s"Illegal port range for tests! $stringRange")
+    val first = limits.head.toInt
+    val second = limits.last.toInt
```
vsuharnikov, line 75: Or by slightly less amount of code:
```scala
val limits = stringRange.split('-').map(_.toInt)
if (limits.length != 2) throw new IllegalArgumentException(s"Illegal port range for tests! $stringRange")
val Array(first, second) = limits
```

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+  def findFreePort(): Either[AvailablePortsNotFound, Int] = {
+    val result = (currentPosition to portsRange.end).find { i =>
+      Using(new ServerSocket(i))(_.getLocalPort).isSuccess
+    }.orElse {
+      (portsRange.start to currentPosition).find { i =>
+        Using(new ServerSocket(i))(_.getLocalPort).isSuccess
+      }
+    }
```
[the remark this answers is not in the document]
vsuharnikov: Also it looks like we don't need to scan ports in `[currentPosition; portsRange.end]` in `orElse` case, because we scanned them. Probably it should be:
```scala
val result = findFreePortIn(currentPosition to portsRange.end)
  .orElse(findFreePortIn(portsRange.start until currentPosition))
```

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  def getBindings(exposedPort: Int): util.List[PortBinding] =
+    getBindings(Seq(exposedPort))
+
+  private def findFreePorts(portNumbers: Int): Either[AvailablePortsNotFound, Seq[Int]] =
+    (0 to portNumbers).foldLeft(Seq.empty[Int].asRight[AvailablePortsNotFound]) { (acc, _) =>
+      acc.flatMap { accValue =>
+        findFreePort().map(_ +: accValue)
+      }
+    }
```
vsuharnikov, line 51: Probably, there is better combinator for this. Doesn't it, @alexander-branevskiy ?

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,81 @@
...
+import scala.util.{Random, Using}
+
+object PortBindingKeeper extends ScorexLogging {
+
+  case class AvailablePortsNotFound() extends Throwable
+
+  private val stringRange = Option(System.getenv("TEST_PORT_RANGE"))
+    .getOrElse(throw new RuntimeException("Please specify the TEST_PORT_RANGE environment variable"))
+
+  private val portsRange = parsePortRange()
```
vsuharnikov, line 21: We don't use `stringRange` later. Probably it'd better to write something like:
```scala
private val portsRange = Option(System.getenv("TEST_PORT_RANGE"))
    .map(parsePortRange)
    .getOrElse(throw new RuntimeException("Please specify the TEST_PORT_RANGE environment variable"))
```
instead.

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,81 @@
...
+import com.github.dockerjava.api.model.Ports.Binding
+import com.github.dockerjava.api.model.{ExposedPort, PortBinding}
+import com.wavesplatform.dex.domain.utils.ScorexLogging
+
+import java.net.ServerSocket
+import java.util
+import scala.jdk.CollectionConverters._
+import scala.util.{Random, Using}
+
+object PortBindingKeeper extends ScorexLogging {
```
vsuharnikov, line 14: Also, it looks like we don't use `Either`, we just throw exceptions which are inside `Left(ex)`. Probably, it'd better to just throw exceptions.

project/ItTestPlugin.scala
```diff
@@ -58,8 +61,30 @@ object ItTestPlugin extends AutoPlugin {
           val logDirectoryValue = (Test / logDirectory).value
           val envVarsValue = (Test / envVars).value
           val javaOptionsValue = (Test / javaOptions).value
+          val (portRangeLowerBound, portRangeHigherBound) = sys.env.get("INTEGRATION_TESTS_PORT_RANGE").map { range =>
+            val limits = range.split('-')
+            if (limits.length != 2) throw new IllegalArgumentException(s"Illegal port range for tests! $range")
+            val first = limits.head.toInt
+            val second = limits.last.toInt
```
vsuharnikov, line 68: Same logic here

project/ItTestPlugin.scala
```diff
@@ -58,8 +61,30 @@ object ItTestPlugin extends AutoPlugin {
...
+            if (first >= second)
+              throw new IllegalArgumentException(s"Illegal port range for tests! First boundary $first is bigger or equals second $second!")
+            (first, second)
+          }.getOrElse(DEFAULT_PORT_RANGE)
+          val tests = (Test / definedTests).value
+
+          // checks that we will not get higher than portRangeHigherBound
+          if (tests.size * PORT_RANGE_LENGTH > portRangeHigherBound - portRangeLowerBound)
```
vsuharnikov, line 76: 👍

project/ItTestPlugin.scala
```diff
@@ -58,8 +61,30 @@ object ItTestPlugin extends AutoPlugin {
...
+            throw new RuntimeException(
+              s"""Cannot run tests;
+                 |They need at least ${tests.size * PORT_RANGE_LENGTH} available ports,
+                 | but specified interval has only ${portRangeHigherBound - portRangeLowerBound}
+                 | """.stripMargin
+            )
 
-          (Test / definedTests).value.map { suite =>
+          tests.zipWithIndex.map { value =>
+            val (suite, i) = value
```
vsuharnikov, line 85: Or just
```scala
tests.zipWithIndex.map { case (suite, i) =>
```

vsuharnikov (changes requested):

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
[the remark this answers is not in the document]
JennaWestenra: I think, we may never check ports from the start of the range, because we provide range for every test, so there may never be that at some point we released ports and we need to take them again.

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def getBindings(exposedPort: Int): util.List[PortBinding] =
+    getBindings(Seq(exposedPort))
+
+  private def findFreePorts(portNumbers: Int): List[Int] =
+    (0 to portNumbers).foldLeft(List.empty[Int]) { (acc, _) =>
+      findFreePort() +: acc
+    }
+
+  def findFreePort(): Int =
```
alexander-branevskiy, line 46: private

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  case class AvailablePortsNotFound() extends Throwable
+
+  private val portsRange = Option(System.getenv("TEST_PORT_RANGE"))
+    .map(parsePortRange)
+    .getOrElse(throw new RuntimeException("Please specify the TEST_PORT_RANGE environment variable"))
+
+  private var currentPosition = portsRange.start
+
+  def getBindings(exposedPorts: Seq[Int]): util.List[PortBinding] = this.synchronized {
```
alexander-branevskiy, line 24: this.synchronized -> synchronized

project/ItTestPlugin.scala
```diff
@@ -8,6 +8,9 @@ import sbt._
 // Separate projects for integration tests because of IDEA: https://youtrack.jetbrains.com/issue/SCL-14363#focus=streamItem-27-3061842.0-0
 object ItTestPlugin extends AutoPlugin {
 
+  private val PORT_RANGE_LENGTH = 50
```
alexander-branevskiy, line 11: PORT_PER_TEST_AMOUNT or sth like this

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/PortBindingKeeper.scala
```diff
@@ -0,0 +1,67 @@
...
+    val result = findFreePorts(exposedPorts.size)
+    log.debug(s"Successfully generated ports: ${result}")
+    exposedPorts.map(new ExposedPort(_)).zip(result.map(Binding.bindPort)).map { ports =>
+      new PortBinding(ports._2, ports._1)
+    }.asJava
+  }
+
+  def getBindings(cmd: CreateContainerCmd, exposedPorts: Seq[Int]): util.List[PortBinding] = {
+    val cmdExposedPorts = cmd.getExposedPorts.toList.map(_.getPort)
+    PortBindingKeeper.getBindings(cmdExposedPorts ++ exposedPorts)
```
alexander-branevskiy, line 35: probably we can omit ```PortBindingKeeper.```

alexander-branevskiy (changes requested):

alexander-branevskiy (approved):

vsuharnikov (approved):

PR #567, opened 2021-06-01
DEX-1174 A better error when a user tries to cancel a pending order

ngadiyak:


dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
```diff
@@ -305,7 +305,7 @@ class AddressActor(
           sender() ! {
             pc.command match {
               case _: Command.PlaceOrder =>
-                error.OrderNotFound(orderId)
+                error.OrderIsPlacing(orderId) // ask Vyatcheslav Suharnikov why do we return an error
```
vsuharnikov, line 308: I meant ask me, we will discuss, then you will add a comment :D

vsuharnikov (changes requested):

vsuharnikov (approved):

PR #568, opened 2021-06-01
DEX-1160 Faster fail by order statuses

ngadiyak:


dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/DexApiSyntax.scala
```diff
@@ -23,7 +24,23 @@ object DexApiSyntax {
...
     def waitForOrderStatus(assetPair: AssetPair, id: Order.Id, status: HttpOrderStatus.Status): F[HttpOrderStatus] =
-      waitForOrder(assetPair, id)(_.status == status)
+      waitForOrder(assetPair, id) { s =>
+
+        def fail() = throw new RuntimeException(s"Expected status $status for order $id, but found ${s.status}")
+        def check(): Boolean = s.status == status
+
+        status match {
+          case Filled | PartiallyFilled => s.status match {
+              case Cancelled | NotFound => fail()
```
vsuharnikov, line 34: I think, we don't need `NotFound` here, because in `BouncingBalancesTestSuite` we have:
```scala
counterOrders.foreach(dex1.api.place)
submittedOrders.foreach(dex1.api.place)

counterOrders.foreach(dex1.api.waitForOrderStatus(_, Status.Filled))
submittedOrders.foreach(dex1.api.waitForOrderStatus(_, Status.Filled))
```

So, it is possible, that an order hasn't yet been placed and we fail checks.

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/DexApiSyntax.scala
```diff
@@ -23,7 +24,23 @@ object DexApiSyntax {
       waitForOrderStatus(order.assetPair, order.id(), status)
 
     def waitForOrderStatus(assetPair: AssetPair, id: Order.Id, status: HttpOrderStatus.Status): F[HttpOrderStatus] =
-      waitForOrder(assetPair, id)(_.status == status)
+      waitForOrder(assetPair, id) { s =>
+
+        def fail() = throw new RuntimeException(s"Expected status $status for order $id, but found ${s.status}")
+        def check(): Boolean = s.status == status
+
+        status match {
```
vsuharnikov, line 32: Let's rename `status` to `expectedStatus` to distinguish `s.status` and `status`.

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/DexApiSyntax.scala
```diff
@@ -23,7 +24,23 @@ object DexApiSyntax {
...
+          case Filled | PartiallyFilled => s.status match {
+              case Cancelled | NotFound => fail()
+              case _ => check()
+            }
+          case Accepted | NotFound => s.status match {
```
vsuharnikov, line 37: We don't need to match an `expectedStatus` with `NotFound`, because:
* It means there is not such order;
* In `waitForOrderStatus` we are waiting for an order. 
So this method is not applicable for `NotFound`.

vsuharnikov (changes requested):

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/DexApiSyntax.scala
ngadiyak: ok

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/DexApiSyntax.scala
ngadiyak: ok

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/dex/DexApiSyntax.scala
ngadiyak: ok

vsuharnikov (approved):

PR #571, opened 2021-06-01
DEX-1100 Added PermitKeepAliveTime

JennaWestenra:


waves-ext/src/main/scala/com/wavesplatform/dex/grpc/integration/DEXExtension.scala
```diff
@@ -31,14 +31,15 @@ class DEXExtension(context: ExtensionContext) extends Extension with ScorexLoggi
...
+    val permitKeepAliveTime: FiniteDuration = context.settings.config.as[FiniteDuration]("waves.blockchain-updates.min-keep-alive")
 
     val bindAddress = new InetSocketAddress(host, port)
     apiService = new WavesBlockchainApiGrpcService(context)
 
     val builder = NettyServerBuilder
       .forAddress(bindAddress)
       .permitKeepAliveWithoutCalls(true)
-      .permitKeepAliveTime(500, TimeUnit.MILLISECONDS)
+      .permitKeepAliveTime(permitKeepAliveTime.length, permitKeepAliveTime.unit)
```
vsuharnikov, line 42: I think we don't need this, because it is our extension and it doesn't depend on Blockchain Updates Extension. Thus we could have own settings

vsuharnikov (changes requested):

vsuharnikov (approved):

PR #580, opened 2021-06-04
DEX-1193 Filter NFT Tokens

JennaWestenra:


vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
```diff
@@ -54,7 +55,8 @@ class AddressActor(
   store: StoreValidatedCommand,
   recovered: Boolean,
   blockchain: BlockchainInteraction,
-  settings: AddressActor.Settings = AddressActor.Settings.default
+  settings: AddressActor.Settings = AddressActor.Settings.default,
+  getAssetDescription: Asset => Option[BriefAssetDescription]
 )(implicit efc: ErrorFormatterContext)
```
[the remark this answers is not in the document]
JennaWestenra: It's unnessesary to use `efc` here, but in AddressActor we use next methods that require an implicit efc:
```
WsOrder.fromDomain(_)
wsAddressState.putOrderFillingInfoAndStatusNameUpdate(remaining, status)
```

so, anyway we need it implicitly. The only difference, I think, I should do is to use `efc` built in `AddressActor` which will use `getAssetDescription` to get asset decimals for these methods.

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/websockets/HasWebSockets.scala
```diff
@@ -94,12 +94,18 @@ trait HasWebSockets extends BeforeAndAfterAll with BeforeAndAfterEach with HasJw
...
+    squash: Boolean = true,
+    onlySquashed: Boolean = false
+  )(expBs: Map[Asset, WsBalances]*)(expOs: WsOrder*): Unit = {
     eventually {
       if (squash) {
-        c.balanceChanges.size should be <= expBs.size
+        if (onlySquashed == false) {
+          c.balanceChanges.size should be <= expBs.size
+          c.orderChanges.size should be <= expOs.size
+        }
```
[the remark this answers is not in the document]
ngadiyak: fixed

ngadiyak (approved):

dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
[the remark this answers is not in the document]
vsuharnikov: Okay, let's keep this until we refactor the code.

waves-ext/src/main/scala/com/wavesplatform/dex/grpc/integration/dto/BriefAssetDescription.scala
```diff
@@ -2,15 +2,16 @@ package com.wavesplatform.dex.grpc.integration.dto
 
 import com.wavesplatform.transaction.assets.exchange.AssetPair
 
-case class BriefAssetDescription(name: String, decimals: Int, hasScript: Boolean)
+case class BriefAssetDescription(name: String, decimals: Int, hasScript: Boolean, isNft: Boolean = false)
```
vsuharnikov, line 5: A new field with a default value is bad for structures, because:
* The default value it is not always appropriate;
* There could be forgotten cases, thus the compiler won't force us to think about each case.

waves-integration/src/main/scala/com/wavesplatform/dex/grpc/integration/dto/BriefAssetDescription.scala
```diff
@@ -2,7 +2,7 @@ package com.wavesplatform.dex.grpc.integration.dto
 
 import com.wavesplatform.dex.domain.asset.Asset
 
-case class BriefAssetDescription(name: String, decimals: Int, hasScript: Boolean)
+case class BriefAssetDescription(name: String, decimals: Int, hasScript: Boolean, isNft: Boolean = false)
```
vsuharnikov, line 5: Same here

vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/db/DbKeys.scala
```diff
@@ -130,13 +130,19 @@ object DbKeys {
         bb.get(name)
         val decimals = bb.getInt
         val hasScript = bb.get == 1
+        val isNft = bb.get == 1
```
vsuharnikov, line 133: Also there could be no more bytes. Consider `false` by default.

vsuharnikov (approved):

dex/src/main/scala/com/wavesplatform/dex/db/DbKeys.scala
```diff
@@ -130,13 +130,19 @@ object DbKeys {
         bb.get(name)
         val decimals = bb.getInt
         val hasScript = bb.get == 1
+        val isNft = Try(bb.get).fold(_ => false, _ == 1)
```
alexander-branevskiy, line 133: we should add a comment here about backward compatibility. sth like "Try here is needed for backward compatibility"

alexander-branevskiy (changes requested):

dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
```diff
@@ -551,27 +557,21 @@ class AddressActor(
   private def scheduleNextDiffSending(): Unit =
     if (wsSendSchedule.isCancelled) wsSendSchedule = context.system.scheduler.scheduleOnce(settings.wsMessagesInterval, self, WsCommand.SendDiff)
 
-  private def mkWsBalances(forAssets: Set[Asset], includeEmpty: Boolean): Map[Asset, WsBalances] = forAssets
+  private def mkWsBalances(forAssets: Set[Asset], includeEmpty: Boolean): List[WsAssetInfo] = forAssets
```
alexander-branevskiy, line 560: >>

dex/src/main/scala/com/wavesplatform/dex/Application.scala
```diff
@@ -241,7 +241,8 @@ class Application(settings: MatcherSettings, config: Config)(implicit val actorS
     storeCommand,
     recovered,
     addressActorBlockchainInteraction,
-    settings.addressActor
+    settings.addressActor,
+    asset => assetsCache.cached.unsafeGet(asset)
```
alexander-branevskiy, line 245: >>

dex/src/main/scala/com/wavesplatform/dex/api/ws/entities/WsAssetInfo.scala
```diff
@@ -0,0 +1,5 @@
+package com.wavesplatform.dex.api.ws.entities
+
+import com.wavesplatform.dex.domain.asset.Asset
+
+case class WsAssetInfo(asset: Asset, balances: WsBalances, isNft: Boolean)
```
alexander-branevskiy, line 5: in case if we consider revert to Map, field "asset" here will be redundant

dex/src/main/scala/com/wavesplatform/dex/api/ws/state/WsAddressState.scala
```diff
@@ -65,23 +72,38 @@ case class WsAddressState(
...
   def clean(): WsAddressState = copy(changedAssets = Set.empty, ordersChanges = Map.empty)
 
+  def checkOptions(assetInfo: WsAssetInfo, options: Set[WsAddressBalancesFilter]): Boolean =
+    if (options.contains(WsAddressBalancesFilter.ExcludeNft))
+      !assetInfo.isNft
+    else true
+
   private def sameAsInPrevious(asset: Asset, wsBalances: WsBalances): Boolean = previousBalanceChanges.get(asset).contains(wsBalances)
+
+  private def toBalancesMap(assetInfo: Seq[WsAssetInfo]): Map[Asset, WsBalances] = assetInfo.map { assetInfo =>
```
alexander-branevskiy, line 97: mkBalancesMap

alexander-branevskiy (approved):

PR #583, opened 2021-06-08
DEX-1208 Add clean-assets CLI command

JennaWestenra:


dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
```diff
@@ -397,6 +413,17 @@ object WavesDexCli extends ScoptImplicits {
...
           .action((_, s) => s.copy(command = Command.CheckConfigFile.some))
           .text("Reports all unused properties from file")
+          .children(
+            opt[String]("dex-config")
+              .abbr("dc")
+              .text("DEX config path")
+              .valueName("<raw-string>")
+              .required()
+              .action((x, s) => s.copy(configPath = x))
+          ),
```
vsuharnikov, line 423: 👍 :)

dex/src/main/scala/com/wavesplatform/dex/db/leveldb/LevelDb.scala
```diff
@@ -42,7 +42,7 @@ object LevelDb {
     override def iterateOver(prefix: Array[Byte])(f: DBEntry => Unit): Future[Unit] = Future(db.iterateOver(prefix)(f))
   }
 
-  def sync(db: DB)(implicit ec: ExecutionContext): LevelDb[Id] = new LevelDb[Id] {
+  def sync(db: DB): LevelDb[Id] = new LevelDb[Id] {
```
vsuharnikov, line 45: 👍

dex/src/main/scala/com/wavesplatform/dex/tool/AssetCacheCleaner.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.wavesplatform.dex.error.Implicits.ThrowableOps
+import org.iq80.leveldb.DB
+
+object AssetCacheCleaner {
+
+  def cleanAssets(dataDirectory: String): ErrorOr[Unit] = Either.catchNonFatal {
+    val db: DB = openDb(dataDirectory)
+    LevelDb.sync(db)
+  }.flatMap(cleanAssets)
+    .leftMap(ex => s"Error during removing assets: ${ex.getWithStackTrace}")
```
vsuharnikov, line 17: Here is a problem: we don't close `db`. Also, I this it'd better to move this function to `WavesDexCli`, because it is a very specific function, that will be used on `WavesDexCli` only. Also it is okay to throw an exception in `WavesDexCli`.

Suggest to:
1. Move this code `WavesDexCli`, e.g. `def withLevelDb(dataDirectory: String)(f: LevelDb[Id] => Unit): Unit`
2. Wrap `db` in `Using`

vsuharnikov (changes requested):

vsuharnikov (approved):

alexander-branevskiy (approved):

PR #586, opened 2021-06-09
DEX-1185 Removed leveldbjni-all

JennaWestenra:


alexander-branevskiy (approved):

project/Dependencies.scala
```diff
@@ -117,6 +117,7 @@ object Dependencies {
   private val julToSlf4j = "org.slf4j" % "jul-to-slf4j" % Version.slf4j
   private val janino = "org.codehaus.janino" % "janino" % Version.janino
   private val kamonCore = kamonModule("core")
+  private val iq80leveldb = "org.iq80.leveldb" % "leveldb-api" % Version.iq80levelDb
```
vsuharnikov, line 120: Why do we need this?

vsuharnikov (changes requested):

project/Dependencies.scala
JennaWestenra: It has an org.iq80 package. If we leave only levelDBJNA dependencies, then we won't get it provided and get compiler errors

vsuharnikov (approved):

PR #591, opened 2021-06-10
DEX-1196 height in debug status

alexander-branevskiy:


dex-it/src/test/scala/com/wavesplatform/it/matcher/api/http/debug/GetSystemStatusSpec.scala
```diff
@@ -11,7 +11,7 @@ class GetSystemStatusSpec extends MatcherSuiteBase with ApiKeyHeaderChecks {
   "GET /matcher/debug/status" - {
     "should return matcher status" in {
       eventually {
-        validate200Json(dex1.rawApi.getSystemStatus) should matchTo(HttpSystemStatus(MatcherStatus.Working, CombinedStream.Status.Working))
+        validate200Json(dex1.rawApi.getSystemStatus) should matchTo(HttpSystemStatus(MatcherStatus.Working, CombinedStream.Status.Working(10)))
```
vsuharnikov, line 14: Or we could just expect `CombinedStream.Status.Working` without any height to make the test faster.

vsuharnikov (approved):

JennaWestenra (approved):

PR #592, opened 2021-06-10
DEX-1193 changed formatting WsAddressSubscribe

JennaWestenra:


dex/src/main/scala/com/wavesplatform/dex/api/ws/protocol/WsAddressSubscribe.scala
```diff
@@ -60,7 +60,7 @@ object WsAddressSubscribe {
       (__ \ "S").format[Address] and
       (__ \ "t").format[String] and
       (__ \ "j").format[String] and
-      (__ \ "b" \ "f").formatNullable[Set[WsAddressBalancesFilter]]
+      (__ \ "b").formatNullable((__ \ "f").format[Set[WsAddressBalancesFilter]])
```
vsuharnikov, line 63: Wondering, what about:
```json
{
  // ...
  "b": {}
}
```
? Will this work?

JennaWestenra: Hmm.. right
I thought that it would be OK, because there is a set, that would be just empty. But it would be OK with the next structure:

```
{
  // ...
  "b": {
       "f": []
  }
}
```

vsuharnikov (approved):

PR #599, opened 2021-06-17
DEX-1216 Remove ForceStartOrderBook

JennaWestenra:


dex/src/main/scala/com/wavesplatform/dex/actors/OrderBookDirectoryActor.scala
```diff
@@ -357,9 +367,10 @@ object OrderBookDirectoryActor {
 
   case class Snapshot(tradedPairsSet: Set[AssetPair])
 
-  case class ForceStartOrderBook(assetPair: AssetPair)
-  case class OrderBookCreated(assetPair: AssetPair)
+  case class CheckOrderBookAvailability(assetPair: AssetPair)
+  case class OrderBookIsAvailable(assetPair: AssetPair)
```
vsuharnikov, line 371: It seems we don't use this.

dex/src/main/scala/com/wavesplatform/dex/actors/OrderBookDirectoryActor.scala
```diff
@@ -357,9 +367,10 @@ object OrderBookDirectoryActor {
...
   case class AggregatedOrderBookEnvelope(assetPair: AssetPair, message: AggregatedOrderBookActor.Message)
+  case class AggregatedOrderBookEnvelopeSent(assetPair: AssetPair)
```
vsuharnikov, line 373: This class is used only in tests. It seems we don't need this class too.

vsuharnikov (changes requested):

vsuharnikov (approved):

dex-it/src/test/scala/com/wavesplatform/it/matcher/api/http/debug/GetOldestSnapshotOffsetSpec.scala
```diff
@@ -30,7 +30,7 @@ class GetOldestSnapshotOffsetSpec extends MatcherSuiteBase with ApiKeyHeaderChec
         mkOrder(alice, wavesUsdPair, BUY, 10.waves, 3.usd)
       ).foreach(placeAndAwaitAtDex(_))
 
-      validate200Json(dex1.rawApi.getOldestSnapshotOffset) should be(-1)
+      validate200Json(dex1.rawApi.getOldestSnapshotOffset) should be(0)
```
vsuharnikov, line 33: Probably we need to wrap this to `eventually`, because a snapshot might not be created in this time

vsuharnikov (approved):

alexander-branevskiy (approved):

PR #612, opened 2021-06-28
DEX-1214 New build for replay tests

ngadiyak:


JennaWestenra (approved):

project/ItTestPlugin.scala
```diff
@@ -69,7 +69,14 @@ object ItTestPlugin extends AutoPlugin {
...
             (first, second)
           }.getOrElse(DEFAULT_PORT_RANGE)
-          val tests = (Test / definedTests).value
+
+          val tests = Option(System.getenv("REPEATED_CI")) match {
+            case None => (Test / definedTests).value
+            case _ =>
+              val runs = Option(System.getenv("REPEATED_CI_RUNS")).getOrElse("0").toInt
+              val suite = System.getenv("REPEATED_CI_SUITE")
+              for (_ <- 0 to runs) yield (Test / definedTests).value.filter(_.name.contains(suite)).head
```
vsuharnikov, line 78: `0 until runs` otherwise we get more runs than it is required

vsuharnikov: Also let's throw an exception if there is no such test. Also there is an opportunity to slightly simplify the code:
```scala
val suiteName = Option(System.getenv("REPEATED_CI_SUITE")).getOrElse(throw new IllegalArgumentException("Specify REPEATED_CI_SUITE"))
val suite = (Test / definedTests).value.find(_.name.contains(suite)).getOrElse(throw new IllegalArgumentException(s"Can't find test *$suiteName*"))
List.fill(runs)(suite)
```

vsuharnikov (changes requested):

project/ItTestPlugin.scala
ngadiyak: fixed

project/ItTestPlugin.scala
```diff
@@ -69,8 +70,19 @@ object ItTestPlugin extends AutoPlugin {
...
+            case _ =>
+              val runs = Option(System.getenv("REPEATED_CI_RUNS")).getOrElse("0").toInt
+              val suiteName =
+                Option(System.getenv("REPEATED_CI_SUITE")).getOrElse(throw new IllegalArgumentException("Specify REPEATED_CI_SUITE"))
+              val suite = (Test / definedTests).value.find(_.name.contains(suiteName)).getOrElse(throw new IllegalArgumentException(
+                s"Can't find test *$suiteName*"
+              ))
+              List.fill(runs)(suite)
+          }
+          val x = List.fill(3)("foo")
```
vsuharnikov, line 85: What is this? ))

vsuharnikov (changes requested):

project/ItTestPlugin.scala
ngadiyak: it was sth i forget to remove )) sry

vsuharnikov (approved):

PR #618, opened 2021-08-16
DEX-1268 Update dependencies

JennaWestenra:


dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/HasToxiProxy.scala
```diff
@@ -27,7 +27,7 @@ trait HasToxiProxy { self: BaseContainersKit =>
 
   protected def getInnerToxiProxyPort(proxy: ContainerProxy): Int =
     toxiContainer.getContainerInfo.getNetworkSettings.getPorts.getBindings.asScala
-      .find { case (_, bindings) => bindings.head.getHostPortSpec == proxy.proxyPort.toString }
+      .find { case (_, bindings) => Option(bindings).flatMap(_.headOption).exists(_.getHostPortSpec == proxy.proxyPort.toString) }
```
JennaWestenra, line 30: New version of testcontainers returns array bindings as `Null` if it was dynamically set by docker. It caused NPE at container's start, because we log every binding at start. So, I wrapped bindings in Option everywhere I found it.

project/Dependencies.scala
```diff
@@ -6,79 +6,80 @@ object Dependencies {
...
     val jwt = "5.0.0"
 
-    val pureConfig = "0.14.0"
-    val allureScalaTest = "2.13.8"
-    val enumeratum = "1.6.1"
+    val pureConfig = "0.16.0"
+    val allureScalaTest = "2.13.9"
+    val enumeratum = "1.7.0"
 
     val scalaPbJson = "0.9.3"
```
JennaWestenra, line 84: Updating causes `waves-integration-it` tests failures (class scalapb/Message not found - wrong versions of scalapb may cause it, so decided to leave it as it is)

dex/src/main/scala/com/wavesplatform/dex/db/leveldb/LevelDbFactory.scala
```diff
@@ -12,6 +12,7 @@ object LevelDbFactory extends ScorexLogging {
 
   lazy val factory: DBFactory = {
     val isTesting = sys.props.get("sbt-testing").isDefined
+    log.info(s"isTesting $isTesting")
```
vsuharnikov, line 15: I think, we don't need this anymore :)

project/Dependencies.scala
vsuharnikov: Let's write a comment about this.

vsuharnikov (changes requested):

dex-it-common/src/main/scala/com/wavesplatform/dex/it/docker/BaseContainer.scala
```diff
@@ -39,6 +39,7 @@ abstract class BaseContainer(protected val baseContainerPath: String, private va
         .getBindings
         .asScala
         .get(new ExposedPort(internalPort))
+        .flatMap(v => Option(v))
```
alexander-branevskiy, line 42: consider flatMap(Option(_))

vsuharnikov (approved):

PR #621, opened 2021-08-18
DEX-1318 Zero fee accounts

alexander-branevskiy:


JennaWestenra (approved):

dex-it/src/test/scala/com/wavesplatform/it/sync/orders/OrderDynamicFeeWithZeroFeeAccountsTestSuite.scala
```diff
@@ -0,0 +1,226 @@
...
+        wavesNode1.api.wavesBalance(bob) shouldBe ib.bobWavesBalance - 10.waves - matcherFee
+        val expectedUsd = AcceptedOrder.calcAmountOfPriceAsset(10.waves, 1.usd)
+        wavesNode1.api.balance(alice, usd) shouldBe ib.aliceUsdBalance - expectedUsd
+        wavesNode1.api.balance(bob, usd) shouldBe ib.bobUsdBalance + expectedUsd
+        wavesNode1.api.wavesBalance(matcher) shouldBe ib.matcherWavesBalance + 2 * matcherFee - tradeFee
+      }
+    }
+
+    "should return reserved balance based on order for zero fee accounts" in {
+      restartWithNewZeroFeeAccounts(Set(alice))
```
vsuharnikov, line 36: It is a some kind of luck, that this works, because we read the queue from the start (same -1 offset) after this restart and create another transactions (because of a different fee) :) 

Suggest to:
1. Create/use a different account for zero fees. 
2. Transfer to it required for tests funds.

An additional advantage is we don't have to restart the matcher.

vsuharnikov (changes requested):

dex-it/src/test/scala/com/wavesplatform/it/sync/orders/OrderDynamicFeeWithZeroFeeAccountsTestSuite.scala
alexander-branevskiy: fxd

vsuharnikov (approved):

PR #631, opened 2021-09-06
DEX-1369 Added TestSuite for check minimum fee for exchangeTx at node

ngadiyak:


waves-integration-it/src/test/scala/com/wavesplatform/dex/grpc/integration/ExchangeTransactionMinimumFeeTestSuite.scala
```diff
@@ -0,0 +1,176 @@
...
+    super.beforeAll()
+    broadcastAndAwait(IssueUsdTx, IssueBtcTx, issueScriptedAmountTx, issueScriptedPriceTx)
+
+    broadcastAndAwait(
+      mkTransfer(alice, scriptedAccount.toAddress, 100.usd, usd),
+      mkTransfer(bob, alice.toAddress, 100.btc, btc),
+      mkTransfer(bob, scriptedAccount.toAddress, 100.btc, btc),
+      mkTransfer(alice, scriptedAccount.toAddress, 1000.waves, Waves),
+      mkTransfer(alice, scriptedAccount.toAddress, 1000.waves, scriptedAmount, 0.007.waves),
+      mkTransfer(alice, scriptedAccount.toAddress, 1000.waves, scriptedPrice, 0.007.waves)
```
vsuharnikov, line 167: We don't need ".toAddress", because we have an implicit conversion

vsuharnikov (changes requested):
Also, we need two additional test cases:
1. Matcher account is scripted and fit 200 complexity, then Node doesn't require an additional fee
2. Matcher account is scripted and don't fit 200 complexity, then Node requires an additional fee

waves-integration-it/src/test/scala/com/wavesplatform/dex/grpc/integration/ExchangeTransactionMinimumFeeTestSuite.scala
ngadiyak: 1) `.toAddress`-- fixed
2) Matcher account is scripted and fit 200 complexity, then Node doesn't require an additional fee  -- if matcher has script with less than 200 complexity the node requires extra fee
3) Matcher account is scripted and don't fit 200 complexity, then Node requires an additional fee  -- added

JennaWestenra (approved):

dex-it-common/src/main/scala/com/wavesplatform/dex/it/waves/MkWavesEntities.scala
```diff
@@ -238,9 +238,22 @@ trait MkWavesEntities {
...
   ): JExchangeTransaction =
-    mkDomainExchange(buyOrderOwner, sellOrderOwner, pair, amount, price, matcherFee, timestamp = timestamp, matcher = matcher).toWavesJ(chainId)
+    mkDomainExchange(
+      buyOrderOwner,
+      sellOrderOwner,
+      pair,
+      amount,
+      price,
+      matcherFee,
+      timestamp = timestamp,
```
alexander-branevskiy, line 252: >>

alexander-branevskiy (approved):

vsuharnikov (approved):

PR #632, opened 2021-09-07
DEX-1379 Possibility of instant sending diff via WS

ngadiyak:


dex/src/main/resources/application.conf
```diff
@@ -685,6 +685,9 @@ waves.dex {
 
     # Timeout of the batch orders cancel
     batch-cancel-timeout = 18 seconds # = 9/10 * actor-response-timeout, should be enough
+
+    # List of accounts which get balance updates via ws instantly
+    realtime-ws-accounts = ["account-public-key-in-base58"]
```
vsuharnikov, line 690: This should be a comment, the code:
```
realtime-ws-accounts = []
```

vsuharnikov (approved):

dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsAddressStreamRealTimeTestSuite.scala
```diff
@@ -0,0 +1,59 @@
...
+import com.wavesplatform.dex.domain.account.KeyPair.toAddress
+import com.wavesplatform.dex.domain.asset.Asset.Waves
+import com.wavesplatform.dex.domain.order.OrderType.{BUY, SELL}
+import com.wavesplatform.dex.model.OrderStatus
+import com.wavesplatform.it.WsSuiteBase
+import org.scalatest.prop.TableDrivenPropertyChecks
+
+import scala.util.Using
+
+class WsAddressStreamRealTimeTestSuite extends WsSuiteBase with TableDrivenPropertyChecks {
```
alexander-branevskiy, line 16: I think this test should do sth like - comparing timestamps & ensuring that delay between messages is much less than ```ws-messages-interval``` (we can set it 1 hour for example in order to make testing easier). it can be skipped this for now and done later.

JennaWestenra, line 16: I think we need to set at least tick in 1hour to be sure that we can receive an update independently of it in eventually (within ~2 minutes).

Also I don't see where we need `TableDrivenPropertyChecks`) And check file for unused imports

JennaWestenra (changes requested):

dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsAddressStreamRealTimeTestSuite.scala
ngadiyak: made address-actor.ws-messages-interval = 1 hour

ngadiyak: removed TableDrivenPropertyChecks

alexander-branevskiy (approved):

dex-it/src/test/scala/com/wavesplatform/it/sync/api/ws/WsAddressStreamRealTimeTestSuite.scala
ngadiyak: and removed ubused imports

PR #638, opened 2021-09-21
DEX-1394 CLI make snapshot method: getting timeout from settings if not specified in args

ngadiyak:


alexander-branevskiy (approved):

dex/src/main/scala/com/wavesplatform/dex/cli/WavesDexCli.scala
```diff
@@ -173,12 +173,15 @@ object WavesDexCli extends ScoptImplicits {
 
   // noinspection ScalaStyle
   def makeSnapshots(args: Args): Unit = {
+    lazy val (_, matcherSettings) = loadAllConfigsUnsafe(args.configPath)
+    val updatedArgs = matcherSettings.cli.defaultArgs.coverEmptyValues(args)
```
vsuharnikov, line 177: We do this for all arguments, see below `OParser.parse(parser, rawArgs, Args())`

dex/src/main/scala/com/wavesplatform/dex/settings/CliSettings.scala
```diff
@@ -24,7 +24,7 @@ final case class DefaultArgs(
     val overrideDexRestApi = if (args.dexRestApi.isEmpty) dexRestApi else args.dexRestApi
     val overrideNodeRestApi = if (args.nodeRestApi.isEmpty) nodeRestApi.getOrElse("") else args.nodeRestApi
     val overrideAuthServiceRestApi = if (args.authServiceRestApi.isEmpty) authServiceRestApi else args.authServiceRestApi
-    val overrideTimeout = if (args.timeout.length == 0) args.timeout else args.timeout
+    val overrideTimeout = if (args.timeout.length == 0) timeout else args.timeout
```
vsuharnikov, line 27: I think this only the reason of the issue

vsuharnikov (changes requested):

vsuharnikov (approved):

PR #642, opened 2021-09-28
DEX-1421 fix exchange tx ts

JennaWestenra:


dex/src/main/scala/com/wavesplatform/dex/model/OrderBookSnapshot.scala
```diff
@@ -21,17 +22,21 @@ object OrderBookSnapshot {
...
+  def fromBytes(bb: ByteBuffer): OrderBookSnapshot = {
+    val bids = OrderBookSideSnapshotCodecs.decode(bb)
+    val asks = OrderBookSideSnapshotCodecs.decode(bb)
+    val lastTrade = bb.get match {
+      case 0 => None
+      case 1 => Some(LastTrade.fromBytes(bb))
+      case x => throw new RuntimeException(s"Can't deserialize Option as $x")
+    }
+    val ts =
+      if (bb.hasRemaining) bb.getLong
```
vsuharnikov, line 37: The problem here: we can't add new fields after this change. I know, we should migrate to protobuf or something else, but with such change we limit our decision in the future. A new feature could have a more priority than a migration to other format.

So, let's use a principle that we use in `lastTrade`: 0 - there is no timestamp, 1 - we have `netxtTxTimestamp`

vsuharnikov (changes requested):

dex/src/main/scala/com/wavesplatform/dex/model/OrderBook.scala
```diff
@@ -59,14 +65,15 @@ case class OrderBook private (bids: Side, asks: Side, lastTrade: Option[LastTrad
     submitted: AcceptedOrder,
     eventTs: Long,
     getMakerTakerFee: (AcceptedOrder, LimitOrder) => (Long, Long),
+    getOrderExecutedTs: (Long, Long) => Long,
     tickSize: Long = MatchingRule.DefaultRule.tickSize
   ): OrderBookUpdates = {
     val events = Queue(OrderAdded(submitted, OrderAddedReason.RequestExecuted, eventTs))
-    if (submitted.order.isValid(eventTs)) doMatch(eventTs, tickSize, getMakerTakerFee, submitted, events, this)
+    if (submitted.order.isValid(eventTs)) doMatch(eventTs, tickSize, getMakerTakerFee, submitted, events, this, getOrderExecutedTs)
```
alxndrmkd, line 72: Not sure, but maybe it should be checked with `orderExecutedTs` instead of  `eventTs`, to prevent invalid exchange transactions?

alxndrmkd (changes requested):

dex-it/src/test/scala/com/wavesplatform/it/sync/txts/MultipleMatchersTxTsTestSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.wavesplatform.dex.api.http.entities.HttpOrderStatus.Status
+import com.wavesplatform.dex.domain.order.{Order, OrderType}
+import com.wavesplatform.dex.it.api.HasKafka
+import com.wavesplatform.dex.it.docker.DexContainer
+import com.wavesplatform.it.MatcherSuiteBase
+import com.wavesplatform.it.tags.DexItExternalKafkaRequired
+
+import java.util.concurrent.ThreadLocalRandom
+
+@DexItExternalKafkaRequired
```
alexander-branevskiy, line 13: redundant here

dex/src/main/scala/com/wavesplatform/dex/model/MatchTimestamp.scala
```diff
@@ -0,0 +1,9 @@
+package com.wavesplatform.dex.model
+
+object MatchTimestamp {
```
alexander-branevskiy, line 3: probably we should add a comment here that it is used for making exchange transactions with strictly increasing timestamp

dex/src/main/scala/com/wavesplatform/dex/model/OrderBook.scala
```diff
@@ -12,7 +12,13 @@ import com.wavesplatform.dex.settings.MatchingRule
 
 import scala.collection.immutable.{HashMap, Queue, TreeMap}
 
-case class OrderBook private (bids: Side, asks: Side, lastTrade: Option[LastTrade], orderIds: HashMap[Order.Id, (OrderType, Price)]) {
+case class OrderBook private (
+  bids: Side,
+  asks: Side,
+  lastTrade: Option[LastTrade],
+  orderIds: HashMap[Order.Id, (OrderType, Price)],
+  nextTxTimestamp: Long
```
alexander-branevskiy, line 20: lastTxTimestamp?

alxndrmkd (approved):

alexander-branevskiy (approved):

dex/src/main/scala/com/wavesplatform/dex/model/OrderBook.scala
JennaWestenra: it is an increased timestamp, so for next tx you will just use it without modifying. So it's nextTxTimestamp)

dex/src/main/scala/com/wavesplatform/dex/model/MatchTimestamp.scala
JennaWestenra: It doesn't increase timestamp, just choose: use old logic (return takerOrderAddedTs), or find `max(takerOrderAddedTs, orderBookTs)`, but this piece of code looks so simple, so I don't think that comment is needed here

vsuharnikov (approved):

PR #663, opened 2021-11-22
DEX-1497 extend ws order info

JennaWestenra:


alexander-branevskiy (changes requested):

ngadiyak (approved):

dex-it-common/src/main/scala/com/wavesplatform/dex/it/api/websockets/HasWebSockets.scala
```diff
@@ -132,9 +132,9 @@ trait HasWebSockets extends BeforeAndAfterAll with BeforeAndAfterEach with HasJw
   protected def mergeWsOrder(orig: WsOrder, diff: WsOrder): WsOrder = WsOrder(
     id = orig.id,
     timestamp = diff.timestamp,
-    amountAsset = orig.amountAsset.orElse(diff.amountAsset),
-    priceAsset = orig.priceAsset.orElse(diff.priceAsset),
-    side = orig.side.orElse(diff.side),
+    amountAsset = orig.amountAsset,
```
alexander-branevskiy, line 135: probably we should add some assert her verifying that orig & diff ```amountAsset, priceAsset, side``` are equal

alexander-branevskiy (approved):

alxndrmkd (approved):

PR #672, opened 2021-12-13
DEX-1502 influxDB to prometheus

JennaWestenra:


alxndrmkd (approved):

alexander-branevskiy (approved):

dex/src/main/resources/application.conf
```diff
@@ -849,10 +849,6 @@ kamon {
       description = "" # in order to fix warnings
       factory = "" # in order to fix warnings
     }
-    influxdb {
```
alexander-branevskiy, line 852: probably Prometheus port should be configured here explicitly

PR #691, opened 2022-02-21
DEX-1562 fix exec events distribution

alexander-branevskiy:


dex/src/main/scala/com/wavesplatform/dex/actors/events/OrderEventsCoordinatorActor.scala
```diff
@@ -119,11 +119,19 @@ object OrderEventsCoordinatorActor {
 
               }
             }
+
           addressActorCommands.added.foreach(addressDirectoryRef ! _)
-          NonEmptyList.fromList(addressActorCommands.executed.toList).map(AddressActor.Command.ApplyOrderBookExecuted(_)).foreach(
-            addressDirectoryRef ! _
-          )
+          addressActorCommands.executed.foldLeft(Map.empty[Address, Vector[AddressActor.OrderBookExecutedEvent]]) { case (acc, event) =>
+            event.affectedAddresses.foldLeft(acc) { case (acc, adr) =>
```
JennaWestenra, line 125: you can use here `foldMap`

JennaWestenra (approved):

dex/src/main/scala/com/wavesplatform/dex/actors/events/OrderEventsCoordinatorActor.scala
alexander-branevskiy: it's possible but I think performance/explicitness with two foldLeft will be better here. (actually, the first version was with foldMap)

alxndrmkd (approved):

PR #718, opened 2022-07-12
DEX-1638 graceful leveldb shutdown - try 1

JennaWestenra:


dex/src/main/scala/com/wavesplatform/dex/Application.scala
```diff
@@ -613,6 +624,9 @@ class Application(settings: MatcherSettings, config: Config)(implicit val actorS
       forceStopApplication(StartingMatcherError)
   }
 
+  private def lookupByNameAndWrap(actorSystem: ActorSystem, name: String) =
+    ExecutorInstrumentation.instrumentExecutionContext(actorSystem.dispatchers.lookup(name), name)
```
alexander-branevskiy, line 628: 1. actorSystem dispatcher is instrumented by default, so we don't need to instrument them twice.
2. Just use native java methods to create executor service cuz we don't need a dispatcher for leveldb
```
  import kamon.instrumentation.executor.ExecutorInstrumentation
  import java.util.concurrent.{Executors, TimeUnit}
  import scala.concurrent.ExecutionContext

  //creation:
  val ec = {
    val se = ExecutorInstrumentation.instrument(Executors.newSingleThreadExecutor(), "leveldb-common-ec")
    ExecutionContext.fromExecutorService(se)
  }

  //for shutting down:
  ec.shutdown()
  ec.awaitTermination(30, TimeUnit.SECONDS)
  ```
  
  use the snippet above as an example

alexander-branevskiy (changes requested):

PR #744, opened 2022-09-13
DEX-1686 address actor failures logging

stavegan:


dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
```diff
@@ -811,7 +825,7 @@ class AddressActor(
       .onComplete {
         case Success(Some(error)) => self ! Event.StoreFailed(orderId, error, command)
         case Success(None) => self ! Event.StoreSucceeded(orderId, command); log.trace(s"$command saved")
-        case _ => throw new IllegalStateException("Impossibru")
+        case _ => throw new IllegalStateException("Unreachable failure")
```
alexander-branevskiy, line 828:
```
case Failure(th) => 
   val exception = new IllegalStateException("Unreachable failure", th)
   log.error(s"$FailurePrefix: Unreachable failure", exception)
   throw exception
```

dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
```diff
@@ -145,7 +145,10 @@ class AddressActor(
         case _ =>
           origActiveOrder match {
             case None => reserve(orderReserve) // Received a partially filled order
-            case Some(origActiveOrder) => throw new IllegalStateException(s"Address already registered $origActiveOrder")
+            case Some(origActiveOrder) =>
+              val exception = new IllegalStateException(s"Address already registered $origActiveOrder")
+              log.error(message = s"$FailurePrefix: address already registered", exception)
```
alexander-branevskiy, line 150: i think we can omit explicit arg ```message``` here -> ```log.error(s"$FailurePrefix: address already registered", exception)```

and for all cases below.

dex/src/main/scala/com/wavesplatform/dex/actors/address/AddressActor.scala
```diff
@@ -868,7 +884,15 @@ object AddressActor {
     settings: AddressActor.Settings = AddressActor.Settings.default,
     isLpAccount: Address => Boolean = _ => false,
     getAssetDescription: Asset => BriefAssetDescription
-  ): Props = Props(
+  ): Props = Props {
+    val log: LoggerFacade = logger.copy(prefix = s"AddressActor[$owner]")
+    val loggedGetAssetDescription: Asset => BriefAssetDescription = asset =>
```
alexander-branevskiy, line 889: 👍

alexander-branevskiy (approved):

PR #751, opened 2022-09-24
DEX-1694 dynamic asset fee config

JennaWestenra:


dex/src/main/scala/com/wavesplatform/dex/caches/OrderFeeSettingsCache.scala
```diff
@@ -1,21 +1,86 @@
...
+      .map(_._2.withDynamicCustomAssets(customAssetsState.get().getAssetsForOffset(offset)))
       .getOrElse(throw new IllegalStateException(s"Order fee settings are not set for offset $offset"))
 
 }
+
+object OrderFeeSettingsCache {
+
+  final case class AssetsActionForOffset(offset: Offset, assets: Set[Asset], isAdded: Boolean)
+
+  final case class CustomAssetFeeState(assetFeeState: Map[Offset, Set[Asset]] = Map.empty) {
```
stavegan, line 46: Probably we should get rid of the default argument here:

```
object CustomAssetFeeState {

  val empty: CustomAssetFeeState =
    CustomAssetFeeState(state = Map.empty[Offset, Set[Asset]])
}
```

